PR #4, opened 2015-04-29
minor amendments to use case text

6a6d74:
As provider of these use cases I suggest a couple of minor changes to the following use cases

4.2 Habitat zone verification for designation of Marine Conservation Zones
4.46 Creation of “virtual observations” from “analysis” phase of weather prediction model

Please MERGE if you're content with these changes.


6a6d74:
@Fransie - thanks!

PR #134, opened 2015-12-17
Minor updates

PBarnaghi:
Minor updates and checked with W3C validator and fixed validation errors.


PBarnaghi:
Minor updates and checked with W3C validator and fixed validation errors.

PR #203, opened 2016-01-12
establishing a stronger relationship with the WG charter

6a6d74:


6a6d74:
@Fransie ... based on discussion in last week's call I have attempted to make a more formal connection between the BP doc and the deliverables stated in the [charter](https://www.w3.org/2015/spatial/charter#bp). Hopefully this addresses your concern.

In particular, see the "Status of this document" section, which now says that we will create a new vocabulary in the future if we deem necessary.

cc/ @PBarnaghi @lvdbrink @philarcher1

PR #238, opened 2016-03-30
first pass,  with placeholders for modularisation and future work

KerryLea:
sections.


KerryLea:
accepting my own mods

PR #262, opened 2016-05-24
Creating static version of the ontology

philarcher:
Essentially getting rid of LODE artifacts, replacing them with static markup. Many of the styles are retained however.

Doc has been put through HTML and CSS validators and the link checker.


KerryLea:
all ready for fpwd (hoping!)

PR #299, opened 2016-07-12
Kjanowicz ssn

kjano:


dr-shorthair:
Too many clashes to resolve automatically, so I made the changes proposed by Jano manually.

PR #320, opened 2016-07-29
Add links from glossary to corresponding terms in ISO/TC 211 Terminology register

dr-shorthair:


6a6d74:
@dr-shorthair ... many thanks. You'll see I added some colour coding to your suggested comparison activities so that readers don't see this as part of the final text. 

FWIW, I checked out your edit branch made changes and pushed to `upstream/simon-bp` to automatically include my edit in this PR. Easy ...

PR #323, opened 2016-07-29
Update sosa.ttl

kjano:
Removed Activity and changed descriptions that made use of it.

Renamed Observer to Host and adjusted description.

Changed SamplingFeature to Sample and changed description.

Added Actuation.

Reworked the relations to fit the new/changed classes (e.g., they were all restricted to observations only).

If you would like to make changes, please start from this version (it is based on discussions around w3c/sdw-sosa-ssn#58 and w3c/sdw#311).


kjano:
@dr-shorthair , @axh599 , is this a good compromise based on the discussions over the last days?

PR #339, opened 2016-08-08
Documentation suggestions

dr-shorthair:
A few suggestions to clarify the documentation. 
Possible changes to property-names: 
- featureOfInterest --> hasFeatureOfInterest
- sample --> sampleOf
  Possible change to definition of 'Observation'

### DIFF
--- a/ssn/rdf/sosa.ttl
+++ b/ssn/rdf/sosa.ttl
@@ -1,5 +1,4 @@
 # baseURI: https://www.w3.org/ns/sosa
-# imports: http://schema.org
 
 @prefix dc: <http://purl.org/dc/elements/1.1/> .
 @prefix meta: <http://meta.schema.org/> .
@@ -12,9 +11,6 @@
 @prefix terms: <http://purl.org/dc/terms/> .
 @prefix xhtm: <http://www.w3.org/1999/xhtml> .
 @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
-
-
-# Meta Properties
 
 ns:license
   rdf:type owl:AnnotationProperty ;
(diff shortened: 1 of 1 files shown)


arminhaller:
Agree to Simon and in the same spirit:

phenomenonTime -> hasPhenomenonTime
resultTime -> hasResultTime

Otherwise looks great, we have something to go back to the group tomorrow!

kjano:
I like the has versions as well but didn't somebody object to them before?

kjano:
Btw, do you also get errors opening the file in topbraid composer? everything works fine but it returns a 'Multiple children of property element' error. I do my edits in sublime (with rdf/ttl highlighting package) but used topbraid today.

arminhaller:
I guess maybe the naming compatibility to SSN may be an issue. SSN seems to be not entirely consistent, for example, using featureOfInterest rather than the hasFeatureOfInterest. We may need to "inherit" that.

One thing I forgot to mention before, although we are currently not doing it anyway. We should definitely not import schema.org (I suggest to even make up our own annotation properties rather than reusing them as we have done initially for the domain/range annotations) in the core for two reasons:
- Simplicity
- Schema.org is not exactly following the principles of ontology engineering, i.e. not importing any prior ontologies, but just copying classes/properties. I don't think we should give them the courtesy to do the reverse, importing schema.org.

dr-shorthair:
> do you also get errors opening the file in topbraid composer?

I did earlier today, but not now. 

Generally find TopBraid's ttl to be very clean, the only issues being 
(i) excessive number of @prefix declarations at the top of the file.,
(ii) gratuitous creation of an owl:Ontology resource, corresponding to the file, which TopBraid uses to support its graph management.  
I'm happy to work around both of those. 

I've just started using Sublime (previously Brackets and before that Ultraedit). Nice.

PR #371, opened 2016-09-08
amended intro section for coverages and glossary def

6a6d74:
… based on suggestions from @dr-shorthair (aka Simon Cox


6a6d74:
@dr-shorthair - thanks for the [suggestions](https://lists.w3.org/Archives/Public/public-sdw-wg/2016Sep/0110.html)

PR #426, opened 2016-12-07
Clarified scope of QB4ST

rob-metalinkage:
PR in small chunks so can be accepted atomically.


dmitry-brizhinev:
Looks good to me

PR #427, opened 2016-12-08
Add information about client application

zhiduozhang:


KerryLea:
will close this -- Duo can we talk about it? I think it needs more than this (and as it was now a while ago there are some minor conflicts, so easiest to handle this way).

PR #455, opened 2016-12-14
Minor changes in SOSA

rgcmme:
- Minor corrections in documentation.
- Unified the use of Procedure (instead of Process) in the ontology.


kjano (approved):
Good clean-up of some naming issues.

PR #459, opened 2016-12-16
SSN - add informative links to preliminary SOSA-O&M and SOSA-om-lite mappings

dr-shorthair:
Since there are RDF files available, we can provide informative links now, so that interested parties can evaluate these in context of SSN/SOSA-OM mappings.


arminhaller:
There seems to be an HTML error in the commit with the &

dr-shorthair:
SSN - Superseded by subsequent work

PR #471, opened 2016-12-19
changed Danh's affiliation

danhlephuoc:


KerryLea (approved):
good -- I even checked the html effect (but this was very painful, would be happy to know a better way. Perhaps I should have cloned danh's branch to check it?)

PR #472, opened 2016-12-20
removed duplication of sections in the beginning of the SOSA section

arminhaller:


danhlephuoc (approved):
There is no conflicts with other sections, removing contents look fine, I'll give it a go then

PR #473, opened 2016-12-20
Restore development that identified change to ssn:sensor

KerryLea:
for the 3rd time.  For the 1st time, slight change to wording.

### DIFF
--- a/ssn/index.html
+++ b/ssn/index.html
@@ -136,7 +136,7 @@ Drawing on considerable implementation experience with SSN since it was first pu
 Specifically,
 <ol>
 <li> The namespace has changed to a permanent W3C address. </li>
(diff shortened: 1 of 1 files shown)


KerryLea:
Now PR includes Notes change commit as well

KerryLea:
For more explanation the second commit " 	Delete Notes added in error" refers to a change that was made during the meeting of the vote in which a minor change was pulled together with an unnoticed significant change without  any discussion.  And for the record, I don't like the change. But this is not the place to discuss it.

KerryLea:
commit mesage was slightly wrong: should be changed every  "modularis" to "modulariz" as required bt W3C rules

arminhaller (approved):

PR #475, opened 2016-12-20
add back readability spacing for 2ontology alignments (4th time)

KerryLea:
there are no tabs but space characters  have been used
Aim to replace by a better alignment using css if possible.


danhlephuoc (approved):
I went through line by line, it looks good for merging

PR #478, opened 2016-12-21
added namespace to SOSA section in response to RESOLUTION: Insert a N…

arminhaller:
…S block at top of SOSA section


philarcher:
This was effectively taken care of in the 5 Jan 2017 publication preparations.

PR #512, opened 2017-01-24
Initial alignment from SSN to SOSA

dr-shorthair:
1. copy ssn.owl to ssn.ttl for tool compatibility
2. create ssn-sosa.ttl containing initial alignment of SSN to SOSA
based on the table https://www.w3.org/2015/spatial/wiki/Mapping_Table


kjano:
An initial alignment between SSN and SOSA.

PR #518, opened 2017-01-29
SOSA - add sosa:Actuation to domain/range of result properties

dr-shorthair:
I noticed that sosa:Actuation had not been included in the domain/range of hasResult/isResultOf properties, where I think it should have been in order that these properties support the general case of observation, actuation or sampling.


kjano:
Oh yes, you are right.

PR #527, opened 2017-02-03
removed closed issue refs from BP4 and added example 

lvdbrink:
removed closed issue refs from BP4 and added example of crawled spatial data


6a6d74:
Hi @lvdbrink ... just checking. Do we know what `2100 Aaron Way, Nanaimo, BC` points to? I wouldn't be totally keen on having my house singled out. Would it make more sense to point to a public building, like Nanaimo Building Inspection at `238 Franklyn St, Nanaimo, BC`?

Don't know if searching this address returns the kind of result you wanted?

Just a thought.

PR #529, opened 2017-02-04
Klt

KerryLea:
-Added missing comments to ssn (specgen needs them)
-removed ssn.owl which duplicates ssn.ttl (somebody copied ssn.owl to ssn.ttl but left ssn.owl behind)


danhlephuoc (commented):
two cosmetic changes on sensor, and observation, then there are new rdfs:comment added to ssn:hasInput, ssh:hasOutput, ssn:hasValue, ssn:observedBy, ssn:startTime. There are no conflicts with the other changes, but, i think i'm not in the position to approve it.

KerryLea:
It appears to me that  I pushed this without  committing the changes that I made after checking with protege plus pellet. These changes are incorporated in the third commit here (i.e adding 2 full stops), but please it is important that these commits here all get merged  as one transaction else ssn will have syntax errors.

KerryLea:
thks

PR #543, opened 2017-02-08
Updated sosa-ssn alignment

dr-shorthair:
I updated the ssn-sosa.ttl graph to reflect refined understanding from the last week's email discussions. 

Note that owl:equivalentClass/Property relations indicate potential opportunities to streamline SSN by adopting the SOSA resources directly, alongside an import of SOSA. 

Included comments for sub-class and sub-property relationships to explain why.

### DIFF
--- a/ssn/rdf/sosa.ttl
+++ b/ssn/rdf/sosa.ttl
@@ -134,7 +134,6 @@ sosa:Result
   skos:definition "The Result of an Observation, Actuation, or Sampling. Such result can, for instance, store an observation's value via the hasValue property."@en ;
   skos:example "The value of 20m as the height of a certain tree."@en ;
 .
-.
 sosa:Sample
   rdf:type rdfs:Class ;
   rdf:type owl:Class ;
--- a/ssn/rdf/ssn-sosa.ttl
+++ b/ssn/rdf/ssn-sosa.ttl
@@ -13,70 +13,80 @@
 @prefix ssn-sosa: <http://www.w3.org/ns/ssn-sosa#> .
 @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
 
+sosa:ObservableProperty
+  rdfs:comment "ssn:Property is super-class of all properties including those that are not necessarily 'observable'. " ;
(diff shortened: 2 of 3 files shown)


maximelefrancois86 (dismissed):
> Not clear how to relate this to ssn:hasValue, which is an owl:ObjectProperty

indeed, `sosa:hasValue` and `oldssn:hasValue` are of different types. 

OWL forbids us to align them in any ways.

Anyways, now that ObservationResult and SensorOutput are merged into Result, shouldn't we get rid of ssn:hasValue altogether ?

maximelefrancois86 (commented):
sosa:ObservableProperty
 rdfs:comment "ssn:Property is super-class of all properties including those that are not necessarily 'observable'. " ;
 rdfs:subClassOf ssn:Property ;

--> OK, conform to the proposed solution for ISSUE-87

maximelefrancois86 (commented):
ssn:FeatureOfInterest
   owl:equivalentClass sosa:FeatureOfInterest ;

--> OK

maximelefrancois86 (changes requested):
If prefix ssn refers to the old ssn
 -> change prefix declaration to http://purl.oclc.org/NET/ssnx/ssn#

Else, if prefix ssn refers to the new ssn
 -> I think we agreed there should be no redefinition of ssn:Xyz when there exists sosa:Xyz ?

dr-shorthair:
The intention of this alignment file was to document the potential equivalentClass/Property and sub-Class/Property relationships. 
A file already exists in GitHub which needs updating to reflect discussions in the last week. 

The URIs/namespaces reflect the **current** state of the ontology in the file ssn.ttl and sosa.ttl. 
If these change in future, then this alignment should be changed, but corresponds with the current state of the repository. 

Lets not get tangled up here - the concerns Maxime expresses are valid, but their resolution belongs in future commits. This pull request should be accepted as-is since it is consistent with the current state of sosa.ttl and ssn.ttl 

As mentioned in the pull-request, where we agree that the relationship is indeed equivalentClass/Property, then there is an option to remove the ssn:Xyz in favour of sosa:Xyz.

PR #549, opened 2017-02-10
Update sosa.ttl

kjano:
Updated definition for FeatureOfInterest to account for actuators.


kjano:
@arminhaller  can you briefly look at this pull request. It just broadens the FOI definition to include Actuators. I need this for the next telco. Thanks.

PR #561, opened 2017-02-20
added voaf:Vocabulary class to ontology instances

maximelefrancois86:


maximelefrancois86:
See 

- https://www.w3.org/2015/spatial/wiki/Voaf_vocabulary
- and  https://lists.w3.org/Archives/Public/public-sdw-wg/2017Feb/0404.html

PR #570, opened 2017-02-21
Update sosa.ttl

kjano:
Addresses the issue where FOI relates to observations and actuations but the isFeatureOfInterestOf and sosa:hasFeatureOfInterest properties only hold for observations. [We may have to do the same for Sampling (https://www.w3.org/2015/spatial/wiki/Sampling) but this remains to be discussed]


maximelefrancois86 (approved):
good to me, ok to merge

kjano:
@arminhaller @maximelefrancois86  The **proposed** additional code is now at: https://www.w3.org/2015/spatial/wiki/Actuation#Actuation_and_Actuators_in_SOSA

PR #578, opened 2017-02-23
corrected sosa:madeByActuator ->  sosa:actuationMadeBy

maximelefrancois86:
patch to meet resolution 4 in the minutes 21/02

### DIFF
--- a/ssn/rdf/sosa.ttl
+++ b/ssn/rdf/sosa.ttl
@@ -225,12 +225,12 @@ sosa:hosts
   owl:inverseOf sosa:hostedBy ;
   skos:definition "Relation between a Platform and a Sensor or Actuator hosted or mounted on it."@en ;
 .
-sosa:madeByActuator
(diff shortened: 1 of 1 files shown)


maximelefrancois86:
see related mail in the list https://lists.w3.org/Archives/Public/public-sdw-wg/2017Feb/0477.html

Resolution 4 in the minutes is: use sosa:madeActuation/sosa:actuationMadeBy
for the property linking Actuation and the Actuator class

After the call, the wrong option was selected in the wiki, and implemented
in sosa.ttl

That is, sosa:madeActuation/sosa:madeByActuator was originally selected in
the wiki as "the option we voted for", and implemented in sosa.ttl.

This is surely just a little mistake, so:
- I updated the wiki in accordance to the resolution;
- I issued a pull request to correct sosa.ttl --
https://github.com/w3c/sdw/pull/578

PR #584, opened 2017-02-28
Update sosa.ttl

kjano:
hostedBy --> isHostedBy as voted for during today's meeting.


maximelefrancois86:
Jano,  can you also change the label for sosa:isHostedBy please ?

and I guess we should relax the definition also, because in its current
state, what I understand is that you need to be either a sensor or an
actuator to be the suject of sosa:isHostedBy.
In SSN, there may be instances of ssn:System that are neither sensors nor
actuators, and that still can be hoted by some platform ?

Best,
Maxime

kjano:
I corrected the label.

PR #585, opened 2017-03-01
Changed sosa:Platform comment to reflect resolutions from today's telco

arminhaller:


arminhaller:
Removed Sampling Devices in the rdfs:comment for now. We don't have the class Sampling Device at the moment in SOSA nor in SSN. Pending resolving of https://www.w3.org/2015/spatial/track/issues/92

kjano:
Personally, I am not in favor of this change but if this is a good first compromise then let us go with it.

dr-shorthair:
ALso see https://www.w3.org/2015/spatial/wiki/Sampling

PR #589, opened 2017-03-01
Resolutions 1, 2, 3, 4, 9 from call 28-02

maximelefrancois86:
Implementation of some of the resolutions in https://www.w3.org/2017/02/28-sdwssn-minutes

- 1. Put VOAF into SSN to make ontology instances voaf:Vocabulary
- 2. Put VOAF into SSN to make ontology instances voaf:Vocabulary
- 3. add vann:preferredNamespacePrefix and vann:preferredNamespaceUri metadata in ssn
- 4. add vann:preferredNamespacePrefix and vann:preferredNamespaceUri metadata in SOSA
- 9. Use owl:versionInfo in SSN


arminhaller:
All good in these commits, but the "owl:versionInfo "New modular version of the SSN ontology independent of DUL."@en ;" in SOSA. It was decided that we do not need a versionInfo in SOSA as it has no prior version. Intention is to keep ontology small. Can you just remove this line, then I can merge.

maximelefrancois86:
Actually the owl:versionInfo metadata is added to the ontology in ssn.ttl only, not in sosa.ttl.
Safe to merge then.

PR #604, opened 2017-03-11
Add reference to OGC API Essentials in BP 11

cportele:


6a6d74:
Hi @cportele - strangely, this PR caused a merge conflict. It's simple enough that I've copied the edits into a new commit that I'm about to make and will merge them from there. So I'm rejecting your PR.

PR #621, opened 2017-03-15
removed BP2 as agreed in bp sub-group call (15-Mar)

6a6d74:
note that the open issue about UoM will be picked up in the
“conclusions” section; to be drafted next sprint.


6a6d74:
@lvdbrink ... I'm going to merge it anyway!

thanks for agreeing to do the merges & WG "please review" email tomorrow!

PR #624, opened 2017-03-16
edited BP1 and BP9

lieberjosh:
Also a small edit to BP3 about engineering coordinate reference systems


dr-shorthair:
> A  model of the sensor optics and platform position may be used later to derive geocentric pixel positions.

should be 

> A  model of the sensor optics, and platform position and orientation, together with transmission path effects (if available), may be used later to derive geocentric pixel positions.

PR #632, opened 2017-03-21
Incorporate EO-QB feedback

qxcv:
Bill Roberts kindly gave some feedback on EO-QB [here](https://lists.w3.org/Archives/Public/public-sdw-wg/2017Mar/0142.html). This PR should address all but one of the issues he raised. The exception is his point on CRSs:

> 5.3 "The working group intends to standardize better properties which allow the use of other CRSs" - does that refer to work on updating GeoSPARQL?  not sure what we'll actually be able to achieve in this area.

I'm not certain what the document was referring to there, so I'll have to leave it to @dmitry-brizhinev to clarify at a later date.


KerryLea:
This refers to the BP  approach  to CRS -- which may be stable enough very recently. I will look at  and adjust accordingly.

PR #633, opened 2017-03-21
ACTION-278: add change history for updates since 2nd PWD until 21st,March

danhlephuoc:
I tracked down the closed PRs back to 5th, January, 2017, I was confused with PR https://github.com/w3c/sdw/pull/572 which contains several changes addressing several issues, I'm not sure I missed something.

### DIFF
--- a/ssn/index.html
+++ b/ssn/index.html
@@ -112,40 +112,38 @@ th, td {
     text-align: left;
 }
 
-</style>
-  </head>
+</style> </head>
   <body>
     <section id="abstract">
-      <p>The Semantic Sensor Network Ontology (commonly known as "SSN" or
-        sometimes "SSNO") is an OWL-2 DL ontology for describing sensors and the
-        observations they make. SSN is published in a
-        modular architecture that supports the judicious use of "just enough"
-        semantics for diverse applications, including satellite imagery, large
-        scale scientific monitoring, industrial and household infrastructure,
-        citizen observers, and the Web of Things. SSN is described and examples
-        of
-        its use are given.</p>
+      <p>The Semantic Sensor Network Ontology (commonly known as
+        "SSN" or sometimes "SSNO") is an OWL-2 DL ontology for describing
+        sensors and the observations they make. SSN is published in a modular
+        architecture that supports the judicious use of "just enough" semantics
+        for diverse applications, including satellite imagery, large scale
+        scientific monitoring, industrial and household infrastructure, citizen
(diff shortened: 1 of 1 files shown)


dr-shorthair (commented):
> Change any mention of owl:inverseOf into schema:inverseOf

I hadn't noticed this, but it makes sense. But presumably only applies to SOSA? 

> Add ObservableProperty and ActuablePropery

*ActuatableProperty*

> Change sosa:madeByActuator to sosa:actuationMadeBy

Hmmm. This becomes inconsistent with comparable properties for observation and sampling.

arminhaller:
Please note that we have not changed the owl:inverseOf in SOSA to schema:inverseOf. There was no such decision made and SOSA still uses owl:inverseOf.

maximelefrancois86 (commented):
> 2. Change any mention of owl:inverseOf into schema:inverseOf

As far as I can remember, the pull request I proposed that was named like this has been accepted, then reverted. SOSA still uses owl:inverseOf.

> 4. Fix minor errors in sosa.ttl and ssn.ttl

Could that be a little more precise, or be deleted ?

> 7. Address naming consitency with isObservedBy/observes,madeBySensor/madeObservation

little typo (consistency). 

> 10.1 Add sosa:Actuation to domain/range of result properties

domain and range ? in sosa :-)

> Rename invokes and invokedBy to madeActuation and madeByActuator

indeed, as Simon said, that's inconsistent with madeBySensor/madeObservation. As shown on figure https://www.w3.org/2015/spatial/wiki/File:20170308-foi-property-object-event.png 

> Add hasSimpleResult instead of hasValue

Should mention there is sosa:hasResult instead of ssn:hasValue, and sosa:hasSimpleResult instead of sosa:hasValue ?

> Add ObservableProperty and ActuablePropery

ActuatableProperty ?

arminhaller:
@danhlephuoc could you please address the issues as above and recommit? Also, please try to use an editor that does not make line break changes so that it becomes clear that the only change that was made has been the change to the "change history".

arminhaller:
As of Maxime's comment, it should be  schema:domainIncludes and schema:rangeIncludes.
We do have sosa:actuationMadeBy and sosa:madeActuation and sosa:madeBySensor and sosa:madeObservation in SOSA. This is not entirely consistent, but due to legacy. And we agreed to leave it like that.
It should be ActuatableProperty as maxime pointed out.

PR #634, opened 2017-03-22
Implemented Procedure changes 

arminhaller:
According to resolution made in https://www.w3.org/2017/03/14-sdwssn-minutes as of Option 1 in: https://www.w3.org/2015/spatial/wiki/Procedure_Process and related to Action item: https://www.w3.org/2015/spatial/track/actions/285

We need to think about what to do with Input and Output. I will raise a separate issue on that.

### DIFF
--- a/ssn/ssn_separated/ssn.ttl
+++ b/ssn/ssn_separated/ssn.ttl
@@ -125,11 +125,11 @@ ssn:deployedSystem rdf:type owl:ObjectProperty ;
                    rdfs:seeAlso "http://www.w3.org/2005/Incubator/ssn/wiki/SSN_Deploy#Deployment" .
 
 
-###  http://www.w3.org/ns/ssn/deploymentProcessPart
-ssn:deploymentProcessPart rdf:type owl:ObjectProperty ;
-                          rdfs:comment "Relation between a deployment process and its constituent processes (has-part)." ;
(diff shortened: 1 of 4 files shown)


ssn/ssn_separated/ssn.ttl
```diff
@@ -815,9 +807,9 @@ ssn:Process rdf:type owl:Class ;
                               owl:allValuesFrom ssn:Output
                             ] ;
             dc:source "http://www.w3.org/2005/Incubator/ssn/" ;
-            rdfs:comment "A process has an output and possibly inputs and, for a composite process, describes the temporal and dataflow dependencies and relationships amongst its parts. [SSN XG]" ;
+            rdfs:comment "A procedure has an output and possibly inputs and, for a composite procedure, describes the temporal and dataflow dependencies and relationships amongst its parts. [SSN XG]" ;
             rdfs:isDefinedBy "http://www.w3.org/ns/ssn" ;
-            rdfs:label "Process" ;
+            rdfs:label “Procedure” ;
```
maximelefrancois86, line 812: there are strange double quote character here

maximelefrancois86 (commented):
Ok apart from minor strange quote character, 
if there isn't anymore mention of "Process", the ok to merge

danhlephuoc (approved):
went through the changes, it's ready to mege

PR #635, opened 2017-03-27
fixes to broken <section> elements etc.

6a6d74:
this fixes:
- the issue with the sections after Conclusions (etc.) not being
Appendices
- the broken Requirements x-ref table

… I also updated the (deleted) BP16 to follow the same shape as the
other deleted BPs.


6a6d74:
@lvdbrink - this fixes two of the outstanding things we need to do before release

PR #651, opened 2017-03-30
Fixes #650 by using madeBySensor to link both Observations with the S…

### LINKED ISSUE #650: Use madeBySensor in example as better practive and to avoid detached Observation
Currently only TempObservation1 is linked to the Sensor using madeObservation; and TempObservation2 is not. 
Linking from the Sensor to its Observations does not seem like a best practice in that it makes streaming of multiple Observations less convenient.
I propose removing the madeObservation triple and linking from each Observation to the Sensor using
sosa-core:madeBySensor examples:TempSensorS4_23;

rivettp:
…ensor (TempObservatipon2 was not previously linked at all).


maximelefrancois86:
ok for me. 
One could also explicit the link between the sensor and each of the observations it made using sosa-core:madeObservation , 
especially because we don't expect that any inference can be made with SOSA-core.

arminhaller:
Please note that the example SOSA instance is in general a bit out of date. Your additional property does make sense, but a thorough review of the example would have to be done anyway.
May I ask how you noticed the example, are you member of the W3C working group?

rivettp:
@arminhaller I am not a member of the working group, I am looking to use SOSA in a project and found the example in the github project as I was attempting to learn how it would be used in practice.

arminhaller:
@rivettp That's great to hear. Can you post the link of where you use SOSA? We will need to collect implementation evidence and it would be great to have yours included.

PR #655, opened 2017-04-03
Action 282 forecasting

maximelefrancois86:
Proposed pull request to deal with Forecasts:

 - a sentence about phenomenonTime and resultTime
 - reference to Forecasting/Forecaster/Forecast are in a separate note, and said to be out of scope of the spec.


dr-shorthair:
That appears to get the balance right,

kjano:
Looks good to me as well. Thanks, Maxime!

PR #657, opened 2017-04-04
Klt eo

KerryLea:


KerryLea:
Have checked compliance with BPs throughout and added references to the relevant BPs in the doc.
In some places the eo-qb sould be a reference for BP, too.

PR #660, opened 2017-04-04
12.2 BP5 Positional accuracy

PeterParslow:
As actioned at the Delft meeting, I have attempted to express more about positional accuracy. As this is my first direct contribution, I'm not sure I've got the tone or format right. That said, do also comment on the actual content too!

Summary:

- Drop encouragement to specify the precision of the data (as the note says, that's misleading anyway)
- Re-draft intended outcome in terms of user fitness-for-purpose (included summary of alternative statements here)
- Added quite a large section in "how" on the various approaches.
- Amended the short list of 'example(s) to be added'
- 'spatialResolutionAsDistance' only really applies to coverages - although it is sometimes (ab)used as a short hand for some statement about the source imagery used in the production of a vector dataset!
- The INSPIRE regulation says nothing about positional accuracy, so conforming to it cannot be an example of a positional accuracy metadata statement; moved it to the more general section - but this makes it an example of a dwbp best practice. But are we OK with an example outside of the 'best practices'? There are other EC Directives that set positional accuracy requirements (e.g. for farm payments, https://marswiki.jrc.ec.europa.eu/wikicap/index.php/Positional_Accuracy - so it would be possible to create an example citing one); a better example might be IHO's S-44 which sets out accuracy requirements for hydropgrahic surveys - so survey datasets could state that they conform to S-44 (at one of four spec levels) - I've tried to create this one (although it doesn't have a URI - I've used the document URL & made up a fragment identifier)
- Added a few points on on 'how to test'
- 

Note: I found this page rather long for collaborative editing; could it be broken up into sections on separate GitHub pages?

### DIFF
--- a/bp/index.html
+++ b/bp/index.html
@@ -1162,10 +1162,20 @@ a:Dataset a dcat:Dataset ;
           used. This can be seen as conformance of data with a "standard" - in this case, a (spatial
           or temporal) reference system. This is how you can describe spatial data quality using
           different vocabularies. We will provide an example in this section. </p>
-        <div class="practice">
+	      <p>For some uses, it may be sufficient to simply state conformance to a published specification</p>
+            <pre class="example" id="ex-geodcat-ap-dataset-conformance-with-specification" title="GeoDCAT-AP specification of a dataset conformance with the INSPIRE Regulation on spatial data and services interoperability">a:Dataset a dcat:Dataset ;
+  dct:conformsTo &lt;<a href="http://data.europa.eu/eli/reg/2010/1089/oj">http://data.europa.eu/eli/reg/2010/1089/oj</a>&gt; .
+
+&lt;<a href="http://data.europa.eu/eli/reg/2010/1089/oj">http://data.europa.eu/eli/reg/2010/1089/oj</a>&gt; a dct:Standard , foaf:Document ;
+  dct:title "COMMISSION REGULATION (EU) No 1089/2010 of 23 November 2010
+             implementing Directive 2007/2/EC of the European Parliament
+             and of the Council as regards interoperability of spatial
+             data sets and services"@en ;
+  dct:issued "2010-12-08"^^xsd:date .</pre>
+	  <div class="practice">
           <p><span id="desc-accuracy" class="practicelab">Describe the positional accuracy of
               spatial data</span></p>
-          <p class="practicedesc">Accuracy and precision of spatial data should be specified in
+          <p class="practicedesc">Accuracy of spatial data should be specified in
             machine-interpretable and human-readable form.</p>
           <section class="axioms">
             <h4 class="subhead">Why</h4>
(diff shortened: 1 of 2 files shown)


6a6d74:
CC/ @lvdbrink 

Thanks @PeterParslow - I'm looking at the new content (rendered via RawGit as it makes it easier to read than the HTML itself). 

One thing I see is that ReSpec throws a warning:

```
Found linkless <a> element with text 'absolute postional accuracy' but no matching <dfn>.
```

You could add the definition for 'absolute positional accuracy' in the glossary ... but squinting more closely, I see that it could just be a typo:

```
s/postional/positional/
```

Free free to make the change and push again :-)

6a6d74:
BTW: 

RawGit [1], and the rendering of your index.html [2]

[1]: https://rawgit.com/
[2]: https://rawgit.com/PeterParslow/sdw/patch-1/bp/index.html

6a6d74:
@PeterParslow ...

Having reviewed your contributions (and many thanks for those!) I suggest a few amendments.

1/ ReSpec is still throwing an error for the 'absolute positional accuracy' definition; suggest modifying [line 3732](https://github.com/PeterParslow/sdw/blob/patch-1/bp/index.html#L3732) from

```
<p><dfn data-lt="Absolute positional accuracy">absolute positional accuracy</dfn>:
```

to

```
<p><dfn>Absolute positional accuracy</dfn>:
```

... ReSpec takes care of capitalisation; because you have the same text in both `@data-lt` and the content on the `<dfn>` element, ReSpec takes these as duplicate definitions.

2/ The **Intended Outcome** section reads like a continuation of the **Why** section. In fact, it might be worth adding the paragraph _to_ the **Why** section. The **Intended Outcome** might be something more like:

> Users can determine whether the spatial data is suited to their intended application

... although I think that that statement is a little too generic; it doesn't mention positional accuracy at all! I like the intent that the outcome is about fitness for purpose.

3/ In Example 12, you have kept the temporal example.  We don't mention temporal accuracy anywhere else - albeit that it's a 'position' just like any other coordinate. I'd be happy if either (a) you deleted it, or (b) added a data snippet to illustrate how one would express this in a machine readable way.

4/ In [line 1235](https://github.com/PeterParslow/sdw/blob/patch-1/bp/index.html#L1235) you mention the _sampling distance_ of `1 000m` - did you mean one thousand metres ... in which case suggest writing `1000 meters` (sadly we're cursed with a need to use American English hence the _wrong_ spelling)

6a6d74:
@PeterParslow - also see suggestion from @KerryLea in her [email](https://lists.w3.org/Archives/Public/public-sdw-wg/2017Apr/0077.html)

PeterParslow:
Any chance of modularising this - at 4000 lines, it's becoming unwieldy to edit!
I tried to make the changes that '6a6d74' (who looks like Jeremy to me!) has suggested - it was painful: scroll 3700+ lines down took several minutes & still left the browser unresponsive. I'll try again later.

6a6d74:
@PeterParslow - yes, @6a6d74 is _Jeremy Tandy_ ... don't ask, it was funny at the time!

I'd prefer not to modularise the document at this point as it adds a deal of complexity to the W3C publication process. Sorry.

I guess that you're editing in GitHub online? Have you tried (or are you able) to clone a local repository, make changes offline and then push these back? Just a suggestion.

Thank you for your perseverance :-)

PeterParslow:
I've cloned the repository, fathomed out the tangled set of branches, and made what I think are the relevant changes. Now I'll try to get my local file back into my personal branch: GitHub tries to commit to patch-1, which I can't .... 

Aha, revert to the web interface & add the amended file to my repository. Then repeat the pull request; maybe that will work?

PeterParslow:
Bizarre that it GitHub diff reckons every line has changed. But not visibly, And not visible to Oxygen diff.

PR #671, opened 2017-04-10
Changes to labels and comments as of Action 303

arminhaller:


ssn/rdf/sosa.ttl
```diff
@@ -153,59 +144,52 @@ sosa:Result
...
+  rdfs:comment "Feature on which Observations may be made, which is intended to be representative of a FeatureOfInterest that is not fully accessible."@en ;
   rdfs:comment "Physical samples are sometimes known as 'specimens'."@en ;
   rdfs:comment """Samples are artifacts of an observational strategy, and have no significant function outside of their role in the observation process. The characteristics of the samples themselves are of little interest, except perhaps to the manager of a sampling campaign.
 
 A Sample is intended to sample some FatureOfInterest in an application domain, so there is an expectation of at least one sampledFeature property. However, in some cases the identity, and even the exact type, of the sampled feature may not be known when observations are made using the sampling features."""@en ;
   rdfs:label "Sample"@en ;
-  skos:definition "Feature on which observations may be made, which is intended to be representative of a FeatureOfInterest that is not fully accessible."@en ;
+  skos:definition "Feature on which Observations may be made, which is intended to be representative of a FeatureOfInterest that is not fully accessible."@en ;
```
kjano, line 155: I would only us the upper case for the class name, not for a collection. I.e., for instance, 'Feature on which an Observation is made' versus 'Feature on which observations are made'. This is how we did it in the entire file before.

ssn/rdf/sosa.ttl
```diff
@@ -153,59 +144,52 @@ sosa:Result
...
+  skos:example "Dividing a field site into quadrants."@en ;
   skos:example "Drawing blood from a patient."@en ;
   skos:example "Drilling an observation well."@en ;
   skos:example "Establishing a station for environmental monitoring."@en ;
   skos:example "Registering an image of the landscape."@en ;
-  skos:example "Seiving a powder to separate the subset finer than 100-mesh."@en ;
+  skos:example "Sieving a powder to separate the subset finer than 100-mesh."@en ;
   skos:example "Selecting a subset of a population."@en ;
   skos:example "Splitting a piece of drill-core to create two new samples."@en ;
   skos:example "Taking a diamond-drill core from a rock outcrop."@en ;
```
kjano, line 182: I would suggest to do the examples in the same style as you did for 'Sampler', namely as text for one skos:example triple.

PR #673, opened 2017-04-10
Actioned changes in BP5 / section 12.2; also tweaked wording on Briti…

PeterParslow:
…sh National Grid (Note in section 8 & section 13.2)

BP5:
- clarify relationship to general data quality; this BP concentrates on positional accuracy, as the only (?) bit that's special to spatial. Noted that the INSPIRE specs don't specify positional accuracy.
- included Kerry's suggested words in 'why' (plus some other examples)

- fix definition of 'absolute positional accuracy' (as per Jeremy's input)
- deleted temporal resolution example (because I don't know how to express it a machine readable way)

Section 8 change - see pull request 666 for discussion: corrected OSGB 1936 to being described as a datum; effectively the same discussion applies in section 13.2 - removed reference to OSGB 1936.

### DIFF
--- a/bp/index.html
+++ b/bp/index.html
@@ -583,7 +583,7 @@ $(document).ready( function() {
       <p>So, it may be that you have information in a projected CRS, rather than global latitude and longitude - what should you do? You can publish data as is in one of these many projected CRS, but you need to tell users which particular CRS is being used. A good directory of Coordinate Reference Systems is maintained by the International Association of Oil and Gas Producers: the <a href="http://www.epsg.org/">EPSG Geodetic Parameter Dataset</a>. </p>
 
       <div class="note">
-        <p>It is common for a CRS to be described by its ESPG code. For example, 2-dimensional WGS 84 (Lat/Long) is <a href="http://epsg.io/4326">EPSG:4326</a>, 3-dimensional WGS 84 (Lat/Long/Elevation) <a href="http://epsg.io/4979">EPSG:4979</a>, Web-Mercator (a global projected CRS used in most Web-mapping applications) is <a href="http://epsg.io/3857">EPSG:3857</a> and OSGB 1936 / British National Grid (a national projected CRS) is <a href="http://epsg.io/27700">EPSG:27700</a>. </p>
+        <p>It is common for a CRS to be described by its ESPG code. For example, 2-dimensional WGS 84 (Lat/Long) is <a href="http://epsg.io/4326">EPSG:4326</a>, 3-dimensional WGS 84 (Lat/Long/Elevation) <a href="http://epsg.io/4979">EPSG:4979</a>, Web-Mercator (a global projected CRS used in most Web-mapping applications) is <a href="http://epsg.io/3857">EPSG:3857</a> and British National Grid (a national projected CRS, based on the OSGB 1936 datum) is <a href="http://epsg.io/27700">EPSG:27700</a>. </p>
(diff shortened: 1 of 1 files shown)


lvdbrink:
@PeterParslow I can't determine what changes you made because the diff is not working and there are merge conflicts. Can you either try again, based on the latest index.html, or give me the changes to add in?

@6a6d74 any other solutions?

PeterParslow:
I was just going to ask about this, given that 10 days have passed since Jeremy noticed the problem. Speaking as an IT old time, my guess is that something in the download/edit/upload loop has replaced all the carriage returns with carriage return/line feed pairs (or vice versa).
Just re-uploading it didn't help - it just created a separate pull request.

Looking more closely, it isn't that: saving the Raw view (https://raw.githubusercontent.com/w3c/sdw/gh-pages/index.html, with Firefox) & comparing with the local copy created by GitHub for Windows, they both have x0Dx0A for each line break.

GitHub for Windows can show the tracked changes, in my local repository. I'll create another pull request from their, to see if that helps. Well, actually GitHub for Windows wouldn't let me create a pull request, but it did let me commit my change to patch-1 (locally), and re-sync - but the sync failed "would overwrite your uncommitted changes. Please discard or commit your changes". The it had me worried, as it said I had no local files - before finding them again.

Has noone else on the team ever had whatever problem this is? Seems like I over estimated my competence with GitHub, having only ever used the online version on reasonable sized files before (TC211 XML projects).

Linda, I'll take you up on your offer, and email you my changed file (actually, I'll drop it in this comment, after zipping
[index.zip](https://github.com/w3c/sdw/files/946432/index.zip)
) 
- as I say, GitHub for Windows can spot the differences, so perhaps you'll have more luck uploading it somehow!

lvdbrink:
There have been other changes since this pull request so I compared the file from your zip with **your** previous version (I think). 

The changes I found are listed below, could you check if this is correct? I have made these changes again in the current draft and will commit them soon. 

_In the intro on data quality (which has been merged with the intro on metadata):_ 
This section is based on the Data Quality section from [[DWBP]] and adds a  best practice specific for spatial data, which concentrates on the accuracy of the positions in the data - how close are they to the actual positions of the real world things?

_Appended to the intro on metadata:_ 
        <p>However, that specification makes no statement about the positional accuracy of the data, 
          so on its own, it is only a useful quality statement for users to whom positional accuracy is not that important.</p>

_In the Why section of former BP5:_ 
Some spatial data applications, such as aircraft navigation, require highly accurate data. For others, such as human navigation, a horizontal accuracy of a few metres is good enough. For yet others, such as overlaying weather forecasts on a map, the map is only giving a general indication of place. If the positional accuracy is published together with the data, the user can determine whether it is appropriate to use for their application. Potentially, this makes existing data more re-usable.

_Removed from former BP5 example list:_ 
                <li>second quarter of the 9th century is <em>approx.</em> 825-850 but could be
                  823-852</li>

_In former BP5 example list:_ 
1 000m  > 1000 meters (but I used American spelling)

_In Glossary:_ 
Removed data-lt="Absolute positional accuracy"

PeterParslow:
Linda,
Thanks for trying to fix this.

I must admit I’ve now lost track of what is being compared to what, and what I might have successfully contributed already. So I’ll list everything that I can see (locally) / remember:

I also made changes related to British National Grid, around line 586 and 3033. This may have got in (somehow) via another pull request (660?). OSGB 1936 is the datum on which BNG (the projected reference system) is based, so at line 586 I suggest “and British National Grid (a national projected CRS, based on the OSGB 1936 datum) is” and 3033, just delete OSGB 1936

I reckon I changed a lot more than that, which I’ve attempted to highlight below (bold text).

lvdbrink:
Thanks for listing it all! The only changes that got lost were the ones concerning OSGB 1936. I made those changes again in my version. Reviewed the rest you listed, everything is there. Must have been in your other PR which I merged succesfully. I think it's all good! 

You can check http://w3c.github.io/sdw/bp/ if you want to be sure it's all in there. 

Will close this PR now!

PeterParslow:
Linda, thanks for unpicking my mess.

PR #674, opened 2017-04-10
added operating and survival ranges to ssnx and ssn

maximelefrancois86:
pull request related to ACTION-308
https://www.w3.org/2015/spatial/track/actions/308


arminhaller:
All good. Can you also make the changes in https://github.com/w3c/sdw/blob/gh-pages/ssn/ssn_separated/ssn.ttl. This is the SSN version that is most current. I am not sure if your last github pull request included changes to that file too. Maybe worth rechecking.

maximelefrancois86:
I would rather clean up ssn_separated/ssn.ttl in another pull request. there is a lot to do there:

for instance the first object property defined in ssn_separated/ssn.ttl is ssn:attachedSystem, that needs to be replaced by sosa:hosts. 

There may be several other little issues, I will check that out.

PR #681, opened 2017-04-11
BP4: correct statement about sitemaps

cportele:
I was try to add @larsgsvensson as a reviewer for the PR, but GitHub did not let me do this...

Lars, can you have a look and see, if the new text is ok or if you would propose any changes.


larsgsvensson:
The proposed text is perfectly fine with me. Thanks, Clemens!

Best,

Lars

PR #684, opened 2017-04-12
SSN - Updates O&M alignment 

dr-shorthair:
To use the *Sampler* and *Sampling* classes and related properties, which were accepted into SOSA subsequent to the original alignment.

Relates to https://www.w3.org/2017/03/14-sdwssn-minutes#x21


arminhaller:
That looks all good to me. Although, I have to admit that I did not check the ISO number mappings are correct. I trust you there.

There are some weird characters in some descriptions (e.g. lt.ul.gt. and lt.i.gt.not etc.). Not sure if this is a problem of the diff tool or if they somehow have appeared through some error in the WYSIWG editor.

dr-shorthair:
Those are in the ISO/TC 211 OWL files which I just cached here. Not my work. 
They are not formally part of the alignment of the documentation - just local copies so I could build the alignment graph. I'll remove them from the repo since they are only confusing.

PR #692, opened 2017-04-13
Bill/covjson status update

BillSwirrl:
@jonblower could you please take a look at this PR - main change is to add a couple of sentences at the start of the Motivation section, as requested in feedback provided at the F2F meeting in Delft, to clarify why the working group is publishing this document.  Following that change we should be good to release as FPWD.


jonblower:
Sorry for delayed response @BillSwirrl - I'm happy with this PR.

PR #696, opened 2017-04-13
[QB4ST] Outline was borked because of missing closing tags

tidoust:
In particular, the outline was wrong starting with the "Requirements" section, and the first appendix appeared as a subsection of last section.

I also updated ReSpec config to display the right logos and use HTTPS links as now required by publication rules.

### DIFF
--- a/qb4st/config.js
+++ b/qb4st/config.js
@@ -23,30 +23,30 @@ var respecConfig = {
       }]
       }],
     wg: "Spatial Data on the Web Working Group",
-    wgURI: "http://www.w3.org/2015/spatial/",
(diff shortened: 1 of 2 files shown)


BillSwirrl (approved):
Francois - many thanks for the tidy-ups.  All looks good to me.

tidoust:
Thanks, not sure what process the group uses for merging pull requests, but I suppose @rob-metalinkage is the one responsible for that one, as editor of the spec.

tidoust:
Note I added a couple more fixes done as I was checking publication rules.

PR #699, opened 2017-04-14
Updated Modularization section, fixed some HTML errors, Updated Change History

arminhaller:


dr-shorthair (changes requested):
1. improved modularization diagram - the OBOE and O&M mappings currently only make reference to SOSA classes and properties. The axiomatization in the SSN ontology is more complete of course, so I guess it would be better to import SSN, but the actual mapping graphs would need to be fixed to match the figure
2. I have updated the SOSA-PROV mapping and also generated a HTML fragment in the same form as the OBOE and O&M mappings. Potential non-normative chapter?

PR #703, opened 2017-04-16
Finalised SSN specification class update in WD

arminhaller:


maximelefrancois86 (commented):
There are still those very strange ^M characters, it's very hard to understand what changed

arminhaller:
according to the SSN file before Maxime's pull request 705

PR #705, opened 2017-04-16
added axioms and local restrictions, simplified, harmonized descriptions

maximelefrancois86:
See also https://lists.w3.org/Archives/Public/public-sdw-wg/2017Apr/0185.html


arminhaller:
the SSN file becomes a moving target now ;-). Will need to incorporate these changes to the SSN specification section tomorrow (my time)

kjano:
same for ssn/ssn_separated/ssn.ttl . We should revert both deletions for the sake of keeping all the links working and the change history

PR #707, opened 2017-04-16
reintroduced DUL alignment into SSNX, now sosa:Observation is dul:Event

maximelefrancois86:
Reintroduced DUL alignment into SSNX, proposed separate file for the alignment between SSN and DUL.

Some comments:

 - Observation is now subClassOf dul:Event

See mail https://lists.w3.org/Archives/Public/public-sdw-wg/2017Apr/0192.html 
Also, I propose that madeObservation, madeBySensor, and  observedValue are all aligned to dul:associatedWith instead of some properties would otherwise make instances of sosa:Observation become instances of dul:Situation (for example, madeObservation was aligned to dul:isObjectIncludedIn, whose range is a Situation.)

 - SensorOutput was aligned to Object, ObservationValue was aligned to Region. They are now both subClassOf Resut. Yet Object and Region are disjoint. I hence propose:
   
sosa:Result rdfs:subClassOf [ owl:unionOf ( dul:Region dul:Object ) ] ;

- Deployment was transitively some kind of Process, and is now more similar to Situation. I hence propose:

sosa:Deployment rdfs:subClassOf [ owl:unionOf ( dul:Situation dul:Process ) ] ;


danhlephuoc (changes requested):
I found a bit confused with the fact that ssnx also imports dul like ssn-dul does. To me, it seems that ssnx has to import ssn-dul as well, otherwise, there need to be justifications why they are separated.

maximelefrancois86:
Would rather make: ssnx import ssn-dul, and that's all

Then ssn-dul can be used to use SSN and DUL together without importing the old SSN

arminhaller:
Agree, ssnx should import ssn-dul, thanks Maxime!

PR #712, opened 2017-04-17
SSN - prov mapping

dr-shorthair:
For ISSUE-53
Alignment of SOSA/SSN with PROV (actually, derivation from). 
All non-normative, but well founded in published papers from Compton et al. and Cox.


dr-shorthair:
Use the html fragment 
   frag_for_sosa-prov-mapping.html
If interested, drop in behind the other alignment chapters.

kjano:
Would be great to have this in the document!

dr-shorthair:
Actually am unclear about one bit of OWL:
Should it be 

sosa:usedProcedure
  rdfs:subPropertyOf [
      rdf:type owl:ObjectProperty ;
      owl:propertyChainAxiom (
          prov:qualifiedAssociation
          prov:hadPlan
        ) ;
    ] ;
.

because sosa:Observationt (etc) is sub-class of prov:Activity, else shoudl it just be 

sosa:usedProcedure
      owl:propertyChainAxiom (
          prov:qualifiedAssociation
          prov:hadPlan
        ) ;
.

I have a similar issue aligning Prov with OWL-Time.

PR #729, opened 2017-04-19
harmonized URIs for alignment modules

maximelefrancois86:
all alignment module URIs:

- now end in sosa/<name of the aligned ont> or ssn/<name of the aligned ont>
- are http instead of https

the corresponding namespaces are hash-based


kjano (commented):
Looks good to me.

PR #737, opened 2017-04-22
Updated the SSN document with new images

rgcmme:
Updated the SSN document with new images


dr-shorthair:
The labels on the relationships are a bit jumbled. 
In most cases they are on the opposite end to what would be expected from following the UML convention. 

In UML the role-name is at the object end. So for example I had expected to see 

[Observation]------------------------------hasFeatureOfInterest ->[FeatureOfInterest]
[Observation]------------------------------madeBySensor ->[Sensor]

which is opposite to how you have drawn it. However, you do have 

[Observation]-------------------------------usedProcedure->[Procedure] 

which is what I expected. 

So they are half-wrong, either way. 

Simon

rgcmme:
In fact, I'm not following the UML notation in the figures at all. I was not the original intention.

The principles followed for labelling properties have been more related to understandability:
1) To include the label near the source of the relationship (just opposite to the UML notation)
2) To try to group together those components more related to a concrete conceptual module (e.g., that's the reason usedProcedure appears inside the procedure module instead of inside the Observation/Actuation/Sampling one)

In any case, if you see a lot of value in labelling the arrows that way, I can try to come up with new figures.

dr-shorthair:
The position of the label should be consistent - in the examples I quoted it is not. 

If we do have double-ended arrows with labels on both ends then I think it would be a mistake to use a convention different to the one already in use by an important part of the user community, ie UML, which is widely used in the OGC.

dr-shorthair:
I'm also concerned that the new figures will be out of reach for the SOSA users - too much detail. The original figures only showed SOSA elements. I think we need both. Colour coding is not enough IMHO.

maximelefrancois86:
I like the diagrams very much although I agree with Simon we could tweak
them a little for the best
Raúl could you share the sources? I could be happy to help do half of the
work here.
What about having 2 versions (SOSA/SSN) of each diagram and show/hide them
the same way we do for the documentation?
Best

rgcmme:
The SOSA-only figures are something that I had in mind. I will update the figures:
* Following a uniform way of labelling properties in the UML style.
* Creating also figures where only SOSA components and restrictions appear.

rgcmme:
The previous changes have been implemented in #748

PR #742, opened 2017-04-25
EO-QB updates and some QB4ST fixes

dmitry-brizhinev:
Addressed most of Kerry's TODO list. Details are below.

Also made some fixes to qb4st - need Rob to take a look at them to see if they are correct.

====== DONE =====

Removed :pixelPerDegree (fixes 1, 4)

Changed crs URI (fixes 2)

Created a qb4st:CRS object for the DGGS (fixes 12)

Checked qb4st for relevant updates - added CRS labels (Fixes 3)

Changed union to separate :dataImageValue and :dataPixelValue properties (fixes 6)

Removed prov:wasDerivedFrom :AGDC (fixes 20)

Changed prov:wasAssociatedWith :GeoscienceAustralia to prov:wasAttributedTo (fixes 21)

Changed names of grid squares from arbitrary "s1" and "s2" to "R000" and "R0000" (fixes 27)

===== NOT DONE COMMENTS ====

9. "QB4ST [qb4st] does not currently have a term for that, but it may in the future."
> It still does not

15. Example 8 led:etmBand      led:latMin led:longMax led:value are not in ANU-LED-example.ttl. They need to be (and they nee to be explained). LEd uses "bounds" instead. Also the example seems to be outofdate wrt QB4st in the ontology?  Also it uses led:Pixel -- shouldn't it be a GridSquare (tile?). 
> For Sam

16. SSN might well change its name and prefix shortly.
> For Kerry

17. Several changes to ssn in the area of example 9. Platform, observes, onPlatform, maybe hasMeasurementCapability. Should add ssn:accuracy.
> For Kerry

22. Example 13: check ssn:Property for change
> For Kerry

24. Example 14. timedomain is a time:Interval that appears to be very short (begins and ends at "2001-10-26T21:32:52) . However, it is normal for satellite imagery to instead be given a (real) time period due to the  time taken to travel over the surface.  I like that the particular :grid Square has only a time instant, but surely the timedomain could be longer?
> I haven't acted on this yet. Sounds like this criticism would also apply to the :time dimension values on the actual data. The way it is now is pleasantly simple and I am reluctant to change it.


rob-metalinkage (approved):
Agree with fixes to qb4st.ttl

PR #750, opened 2017-04-25
implemented ACTION-337 and ACTION-342, see minutes of meeting 2017-04-26

maximelefrancois86:


arminhaller:
Looks good. Can you please also include an update to the Change History or make sure that the right names are used in the Change History.

PR #751, opened 2017-04-26
Removed equivalence relation from oldssn:Observation to sosa:Observation and generalised comment on Property, hasProperty and isPropertyOf

arminhaller:


dr-shorthair:
1. Should we perhaps use the term 'SpatialThing' instead of, or in addition to 'entity' - for consistency with the BP recommendations? 
2. ISO/DIS 19156 should be ISO 19156:2011 (it became an IS in 2011, no longer a DIS). "ISO 19156:2011" is the citation form preferred by ISO

arminhaller:
Re 1: We can, but would need to say spatial thing. We do use the word entity in several places, though, and for some of the subproperties spatial thing would be quite far fetched, e.g. systemCapability (accuracy) which may refer to other qualities than spatial
Re 2: I actually did not change anything there, this is just my tool removing whitespaces at the end of a line. I would suggest another Pull request from you, if there are errors in the ISO mapping.

dr-shorthair:
1. See http://w3c.github.io/sdw/bp/#spatial-things-features-and-geometry 
2. OK - I don't think it is the mapping that is a problem, just the citation.

arminhaller:
Re 1: Yes, quote: "the term “feature” is also commonly used to mean a capability of a system, application or component." In the ontology we use Property also for capabilities of a system. Therefore, IMHO we shouldn't use spatial feature.
Re 2: Ok, I can do that in another pull request, or you do it, then I don't have to create another branch?

dr-shorthair:
ISO citation issue has been resolved in https://github.com/w3c/sdw/pull/752

arminhaller:
Somehow the sampling commit has created a conflict with this request, strange?

dr-shorthair:
I guess it is because they both changed the lines in ssnx.rdf which ended with [O&amp;M - ISO/DIS 19156] 
but made *different* changes.

PR #752, opened 2017-04-26
SSN samplingresult

dr-shorthair:
Removed isSamplingResultOf and hasResult

Added constraints to Sampling/Sample to require the Result to be a Sample
Adjusted domain/rangeIncludes where necessary

http://www.w3.org/2015/spatial/track/actions/341


dr-shorthair:
Will look at the document next

PR #755, opened 2017-04-26
Added contributors https://www.w3.org/2015/spatial/track/actions/336, switched Armin with Kerry in Editor list, https://www.w3.org/2015/spatial/track/actions/332

arminhaller:
Couldn't find a way to have line breaks in the Contributor list key/value pair

Switched editor as of decision in https://www.w3.org/2017/04/18-sdwssn-minutes


dr-shorthair:
Have a look at https://github.com/w3c/sdw/blob/gh-pages/time/config.js to see how I did it for OWL-Time. 
Not necessarily ideal, but ReSpec doesn't appear to have an explicit tag for contributors https://github.com/w3c/respec/wiki

PR #760, opened 2017-04-27
BP: Added GeoRSS bib entry

andrea-perego:
GeoRSS is referenced from the section on spatial data vocabularies, and in the revised version of the tables in Appendix A (relevant pull request yet to be created).


6a6d74:
@andrea-perego ... small change merged :-)

PR #780, opened 2017-05-02
Update index.html

kjano:
Should be sosa:Platform  subclass of dul:Object, not PhysicalObject, correct? After all, we decided to have full support for virtual and simulated data, events, a nd deployments.


kjano:
@maximelefrancois86 , I guess this should also trigger 'oldssn:Platform equivalent class  sosa:Platform' to become 'oldssn:Platform subclass  sosa:Platform' , would you agree?

dr-shorthair:
Yes - Platform could be in the cloud.

PR #789, opened 2017-05-04
Add spatial relation / operator to the glossary

cportele:


6a6d74:
@cportele - thank you & merging now.

PR #794, opened 2017-05-04
BP: Change order of BPs in bullet list (editorial)

cportele:
Change order of BPs in bullet list in chapter 9 to increasing
number/importance


6a6d74:
thanks @cportele - merged!

PR #820, opened 2017-05-06
Remove glossary link in <a/> element

cportele:
… because that breaks the link


6a6d74:
@cportele ... Thank you! Merging now.

PR #821, opened 2017-05-06
"dereference" instead of "resolve"

cportele:
Webarch: "Agents may use a URI to access the referenced resource; this
is called dereferencing the URI."

"Resolving a URI" is not really used in the IETF/W3C documents.


6a6d74:
@cportele ... thanks again! Merging now.

PR #824, opened 2017-05-06
Add BP3 in chapter 11

cportele:
In 12.1.3 we say that "the widespread use of links within data is
regarded as one of the most significant departures from contemporary
practices used within SDIs."

To be consistent, we should include BP3 in the list of key BPs in the
"Why are traditional Spatial Data Infrastructures not enough?" section.


6a6d74:
@cportele - another good spot. Thank you.

PR #825, opened 2017-05-06
BP4: Update classification of applications based on Delft F2F

cportele:
In Delft we agreed that SDIs are not applications of spatial data, but
spatial analytics is.


cportele:
I see now that BP4 has more text on each of the classifications, so that change needs more work in the "Possible Approach to Implementation" section. I will have a look at this (tomorrow).

6a6d74:
@cportele : OK & many thanks!

PR #835, opened 2017-05-06
Revision to BP5 (formerly, BP8)

andrea-perego:
BP5 revised to take into account the topics addressed in [BP6](https://w3c.github.io/sdw/bp/#multiplegeometries) (#802).


6a6d74:
Hi @andrea-perego - those changes look good; I'll merge so that folks can include it in their general review of the doc. Many thanks!

PR #841, opened 2017-05-07
Fix reference after removal of "how to use" section

cportele:


6a6d74:
@cportele 

I'll merge this & then make another change ... ReSpec is clever enough to fill in the text for a <a> element. Also, we're using `class="sectionRef"` to style these links.

cportele:
@6a6d74 Oh thanks, when we are done I will have understood ReSpec ;)

6a6d74:
They keep changing it! And I only understand the bits I use :-)

6a6d74:
See PR #842 

```
<a href="http://w3c.github.io/sdw/bp/#bp-exposing-via-api">section 12.3 Spatial data access</a> 
```

becomes:

```
<a href="#bp-exposing-via-api" class="sectionRef"></a>
```

cportele:
I see - will use that in future. Thanks!

PR #850, opened 2017-05-08
Several terms marked Features at risk as of https://www.w3.org/2015/spatial/track/actions/346

arminhaller:


dr-shorthair:
Looking at this set of Features at Risk, I note that they are all (?) related to sensor description details. I would suggest there is strong case for moving them all into a horizontal module (package).

kjano:
@dr-shorthair, very, very good idea! [Which would also make a stronger argument for horizontal modules in the first place]

arminhaller:
Agree. For now, though, we need to mark them as Feature at Risk so that we have the potential to do that in the implementation evidence phase. At least that is my understanding of the process.

kjano:
And the horizontal module could be non-normative meaning that we can have all the classes and properties there without the need for implementation evidence.

dr-shorthair:
Yes - mark as at-risk, but package into a separate graph to make it easier to use or exclude (and also could be documented in a separate chapter that could be marked 'non-normative' like the sample relationships chapter).

dr-shorthair:
Or just do what Jano said right now - move documentation into non-normative chapter.

arminhaller:
I agree, but maybe we need to do that step by step. We are not yet in the implementation evidence phase, so I would leave the terms for now where they are and then move them.

rgcmme:
Typo in the change log: "#SSNhasSurvivalRang" -> "#SSNhasSurvivalRange"

PR #864, opened 2017-05-08
updated changes section

lvdbrink:


lvdbrink:
@6a6d74 could you check if I forgot something?

PR #870, opened 2017-05-08
Distinguish RDF and XML namespaces and their prefixes

### LINKED ISSUE #869: BP: namespace section limited to RDF namespaces?
The namespace section introduced with PR #828 is misleading. It lists for the prefix "gml" the namespace IRI "http://www.opengis.net/ont/gml#" and claims that this namespace would be defined in the GML standard. This is incorrect.  

Almost everyone dealing with GML data or schemas would expect "http://www.opengis.net/gml/3.2" (or "http://www.opengis.net/gml") as the namespace IRI as it is defined in the GML standard. The prefix "gml" is used in the document as the abbreviation for "http://www.opengis.net/gml/3.2" at least in examples 15/16/29. The current section 5 seems to use the term "namespace" with an RDFS bias.

It is GeoSPARQL that specifies "http://www.opengis.net/ont/gml#" and uses "gml" as the abbreviation.

Possible solutions:
* split "5 namespaces" into two sections "5.1 RDF namespaces" and "5.2 XML namespaces" (the prefix "gml" would occur in both);
* list both the RDF and XML namespaces in the table in section 5, but clearly identify them.

My preference would be the first option to separate the different types of usages of the term "namespace".

Note that other XML namespace prefixes are used in the document, too. At least: "georss", "bag", "wml2", "xlink", "sam", "sams".

cportele:
closes #869

But see also the ongoing discussion in that issue.

### DIFF
--- a/bp/index.html
+++ b/bp/index.html
@@ -885,19 +885,9 @@ function init() {
             <td><a href="http://bag.basisregistraties.overheid.nl">Dutch Government Base Registry Adressen en Gebouwen (BAG)</a></td>
           </tr>
           <tr>
-            <td><code>dc</code></td>
-            <td>http://purl.org/dc/terms/</td>
-            <td><a href="#bib-DCTERMS">Dublin Core Metadata Initiative (DCMI) Metadata Terms</a></td>
-          </tr>
-          <tr>
             <td><code>dcat</code></td>
             <td>http://www.w3.org/ns/dcat#</td>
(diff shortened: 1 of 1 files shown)


6a6d74:
For info: you can review what this change looks like via [rawgit](https://rawgit.com/w3c/sdw/bp-edit-clemens-ns/bp/index.html#namespaces)

PR #872, opened 2017-05-09
Fixed syntax bug in namespace table

andrea-perego:
There was an unclosed &lt;tr&gt; tag in the wrong place, and a missing &lt;tr&gt; tag in another place.


cportele (approved):

cportele:
Thanks, Andrea!

PR #874, opened 2017-05-10
BP12: Example about the GeoDCAT-AP API

andrea-perego:
The example about the GeoDCAT-API was originally included in the narrative (see last paragragh in [the relevant wiki page](https://www.w3.org/2015/spatial/wiki/BP_Narrative_2#.284.29_Publish_details_of_fixed_assets_.28e.g._dikes_.26_dams.2C_buildings.2C_roads.2C_critical_infrastructure_etc..29_and_topographical_features_.28e.g._water_bodies.29)).

The proposal is to add it to [BP12 ("Expose spatial data through 'convenience APIs'")](http://w3c.github.io/sdw/bp/#convenience-apis), under point (1) ("Reuse your existing spatial data infrastructure"), after Example 49.

The relevance for that section concerns the fact that the GeoDCAT-AP API allows geospatial catalogues and CSWs to serve geospatial metadata in HTML+RDFa and RDF, by using standard CSW query parameters - namely, `outputSchema` and `outputFormat`, and via HTTP content negotiation. (More details are available in [the API documentation page](https://webgate.ec.europa.eu/CITnet/stash/projects/ODCKAN/repos/iso-19139-to-dcat-ap/browse/api)).

### DIFF
--- a/bp/index.html
+++ b/bp/index.html
@@ -3528,11 +3528,6 @@ Content-type: application/geo+json
                   of writing generalized SPARQL queries and understanding the underpinning 
                   data model.</p> 
               </aside>  
-              
-<aside class="example" id="ex-geodcat-ap-api">
-<p>As far as metadata are concerned, the experimental <a href="https://webgate.ec.europa.eu/CITnet/stash/projects/ODCKAN/repos/iso-19139-to-dcat-ap/browse/api">GeoDCAT-AP API</a> allows data publishers to serve [[ISO-19115]] records in different <a>RDF</a> serialization formats, [[HTML-RDFa]] included, on top of a geospatial catalog, by using the standard [[CSW]] query interface, and supporting HTTP content negotiation.</p>
-</aside>
-              
               </li>  
               
               <li>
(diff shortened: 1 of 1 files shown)


cportele:
I have a concern about this change: The API is an API for metadata records, not for spatial things, which is what BP12 is about. This is also why CSW is not mentioned in the "spatial data access" section.

Maybe the example could be added in the "spatial metadata" section as an example for an API to access metadata records consistent with the guidance in BP12?

andrea-perego:
On Wed, May 10, 2017 at 12:04 PM, Clemens Portele <notifications@github.com>
wrote:

> I have a concern about this change: The API is an API for metadata
> records, not for spatial things, which is what BP12 is about. This is also
> why CSW is not mentioned in the "spatial data access" section.
>
> Maybe the example could be added in the "spatial metadata" section as an
> example for an API to access metadata records?
>

Thanks, @cportele . I'm happy with your proposal. In such a case, the example can be included in BP13, after Example 53 <http://w3c.github.io/sdw/bp/#ex-crowd-sourced-data-metadata>, or just after the last bullet point of section "Possible approach to implementation" ("Publish metadata in both machine- and human-readable format, following [DWBP] Best Practice 1: Provide metadata.").

cportele:
@andrea-perego 
Yes, that should work and nicely addresses the "machine- and human-readable format" part.

andrea-perego:
Thanks, @cportele . Just revised the PR accordingly (see 22f4e87be78a9dee080c2075e477e2e90a1b2d77). Now the example is in BP13.

andrea-perego:
@cportele , about this point:

> I have a concern about this change: The API is an API for metadata records, not for spatial things, which is what BP12 is about. This is also why CSW is not mentioned in the "spatial data access" section.

I wonder whether it would be useful to add a note in BP12, at the end of the "Why" section, explaining that BP12 is about APIs for "data" only, whereas, as far as metadata are concerned, the relevant BP is BP13.

cportele:
@andrea-perego - no problem with that. I think I would add this early in the section 12.3 "spatial data access" as the note applies to the whole section. As it is an informative edit I think we could always do this in the next iteration, too.

PR #879, opened 2017-05-10
[BP] Fix a few markup errors to please pubrules

tidoust:
Also adjusted encoding issues for some author names (Unicode Normalization Form C should be used), and updated links that yield a 301 permanent redirect to avoid the redirect.

### DIFF
--- a/bp/bpconfig.js
+++ b/bp/bpconfig.js
@@ -20,7 +20,7 @@ var respecConfig = {
     }, {
 			name: "Payam Barnaghi",
 			company: "University of Surrey",
-			companyURL: "http://www.surrey.ac.uk/"
+			companyURL: "https://www.surrey.ac.uk/"
 		}],
     otherLinks: [
     {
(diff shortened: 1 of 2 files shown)


andrea-perego:
@tidoust , I think there are still some HTML validation issues - see:

https://validator.w3.org/nu/?doc=http%3A%2F%2Fw3c.github.io%2Fsdw%2Fbp%2Findex.html

tidoust:
@andrea-perego Right but note you're testing the source spec, here, not the generated spec. The errors disappear with the generated spec, except the empty <tr> which I had indeed missed and which I had to fix when I prepared the document to publish. I'll create another PR.

andrea-perego:
Thanks, @tidoust . Yes, I was not actually referring to the warnings.

BTW, please note that, besides the empty &lt;tr&gt;, there are two other errors:

> Error: Duplicate attribute class.
> At line 4350, column 53

> Error: Non-space character in page trailer.
> From line 4587, column 1; to line 4587, column 8

PR #883, opened 2017-05-12
Added a figure with an overview of the SOSA ontology

rgcmme:
Added a figure with an overview of the SOSA ontology.


kjano:
I would just have a tiny suggestion, namely moving 'hosts' to the left and a bit down to improve readability.

PR #895, opened 2017-05-17
mark as outdated

kjano:
Would this be a solution for the moment:

# ---------------------------------------------------------------------------------------------------------------
# OUTDATED, PLEASE USE THE SOSA FILE AT https://github.com/w3c/sdw/blob/gh-pages/ssn/integrated/sosa.ttl INSTEAD.
# ---------------------------------------------------------------------------------------------------------------


maximelefrancois86:
Nice one :-) can you duplicate the change also in the SSN_separated folder ?

PR #912, opened 2017-06-06
Fixes to DUL Alignment

kjano:
Please do not accept this request before the meeting today.


maximelefrancois86:
do the same in the .ttl ?

arminhaller:
Please also add a Change History

PR #916, opened 2017-06-08
Removed/Relaxed axioms 

arminhaller:
Based on the request to allow generic modelling of Observations, Actuations, Sampling and FeaturesOfInterest as of https://www.w3.org/2015/spatial/wiki/Wide_Review

### DIFF
--- a/ssn/index.html
+++ b/ssn/index.html
@@ -543,7 +543,7 @@ their usage are given.</p>
                       <span about="sosa:Observation" rel="rdfs:subClassOf" resource="[_:SOSAObservation1]"></span>
                       <span about="[_:SOSAObservation1]" typeof="owl:Restriction"></span>
                       <span about="[_:SOSAObservation1]" rel="owl:onProperty" resource="sosa:madeBySensor"><a href="#SOSAmadeBySensor">sosa:madeBySensor</a></span>
-                      <span about="[_:SOSAObservation1]" property="owl:cardinality" content="1" datatype="xsd:nonNegativeInteger"><em><b>EXACTLY 1</b></em></span>
+                      <span about="[_:SOSAObservation1]" property="owl:minCardinality" content="1" datatype="xsd:nonNegativeInteger"><em><b>MIN 1</b></em></span>
                       <br>
                       <span about="sosa:Observation" rel="rdfs:subClassOf" resource="[_:SOSAObservation2]"></span>
                       <span about="[_:SOSAObservation2]" typeof="owl:Restriction"></span>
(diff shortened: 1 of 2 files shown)


dr-shorthair (commented):
Disagree with this proposal. The cardinality=1 restrictions are intrinsic to the definition of acts of observation/actuation/sampling - they involve one procedure, concern one property and one foi. Those may have relationships with other, usually more general instance of the same type, but the cardinaility correct.

kjano:
In a nutshell, we want to foster common models without restricting potential encodings for them. Wrt observable properties, we want there to be one temperature, one air pressure, etc, so that we foster interoperability and discovery, but we do not want to make statements about how to encode these, e.g., whether they should be singleton classes, i.e., subclasses of observableProperty, single instances of observableProperty, or instances introduced by way of OWL2 punning. The difference is simply that we cannot enforce this on the SOSA level but do so on the SSN level.

My understanding of the axioms below , e.g. sosa:Observation subClassOf sosa:observedProperty EXACTLY 1, was different, namely that we do not allow for composite observations such as one observation observing air pressure and temperature at the same time. In this sense, observations are atomic. Similarly, observations cannot be shared across sensors, i.e., one observation only results from one sensor and composite observations or the application of opaque post-processing are not intended. Multiple sensors can, of course, work jointly to generate observation *collections*.

dr-shorthair:
> ... we do not allow for composite observations such as one observation observing air pressure and temperature at the same time. In this sense, observations are atomic. Similarly, observations cannot be shared across sensors, i.e., one observation only results from one sensor ... 

Unless a composite observable/sensor is defined. This is quite natural for some quantities which are intrinsically vector or tensors (position, stress & strain, also multi-component/spectral observations), but should probably be avoided for arbitrary combinations of properties. In OMXML we had 'weather' in some of the examples, defined as a combination of temperature, humidity, wind speed and direction. While this is OK in principle, in practice it can get messy as the composite result must be encoded as a sequence or structure in which order counts.

kjano:
Yes, absolutely. You can have observations of _weather_, e.g., to state that it was cold. What you cannot have (in SSN) right now (and IMHO for good reasons) is a single, i.e., the same, observation for all the individual components of _weather_ such as that observation1 would be about multiple (different) observable properties. You can state (using both SOSA and SSN) that 'weather' is an observed property   and 'cold' is a (simple) result.

dr-shorthair:
I think I disagree on this. 

Weather can be 'cold, wet and windy'. In principle this is supported by a composite property and matching composite sensor. In practice this is tricky to encode. 

But things like position, velocity, stress are vector/tensor quantities, where a single component is meaningless without the others and the definition of their basis vectors. So in these cases we must have a multi-component property and sensor. 

Another case for multi-component properties is radiance with is usually observed over a spectrum of wavelength channels.

kjano:
Yes, absolutely, you can have a 'cold, wet and windy' result for weather using a single observation.

dr-shorthair:
FWIW - for those of you who haven't encountered this, the state of stress in a solid (such as at locations in the earth's subsurface) is a tensor quantity, requiring nine numbers to describe (though only six are independent) - see https://en.wikipedia.org/wiki/Cauchy_stress_tensor

dr-shorthair:
BTW - the changes here appear to be exactly the same as 920.

maximelefrancois86:
I suggest we close this pull request, and issue a new one that relaxes only
 
ssn:isPropertyOf a owl:FunctionalProperty
ssn:hasProperty a owl:InverseFunctionalProperty.

This is very important if we want to allow a generic property such as "Temperature" to apply to multiple feature of interest.

Then, I propose to write a last section in http://w3c.github.io/sdw/ssn/#common-modeling-questions that reports on the different modeling possibilities as in https://www.w3.org/2015/spatial/wiki/Generic_vs_Specific_Modeling

will work on this as soon as this pull request gets closed.

PR #920, opened 2017-06-12
Update to SSN Usage

arminhaller:
as of Action-368

Careful, before accepting this Pull request we may need to reject Pull request 916

### DIFF
--- a/ssn-usage/index.html
+++ b/ssn-usage/index.html
@@ -7,19 +7,19 @@
       <meta content="text/html; charset=UTF-8" http-equiv="content-type">
       <script class="remove" src="http://www.w3.org/Tools/respec/respec-w3c-common"></script>
       <script class="remove" src="config.js"></script>
+      <link rel="stylesheet" type="text/css" href="style.css">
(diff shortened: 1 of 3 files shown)


dr-shorthair:
I disagree with these proposed changes. 
The strict cardinality expectations (one Observation has one feature of interest, one observed property, one procedure and one sensor) are an important part of the model. If multiple values are allowed for any of these, then we no longer have a unique mapping from one result to the foi/property/procedure/sensor  . 
As discussed last week, the onus should be on a data provider to ensure that their data has sensible entailments, and if a data user needs to relax one of these for discovery or other purposes, then this should be done outside the data instance.

arminhaller:
Please do not look at the changes to the SSN index.html. I can't really avoid that this change is included, since the other Pull request is not rejected yet (unless I missed some trick in Git). This pull request should only be about the update to index.html of the SSN-Usage document.

maximelefrancois86:
Armin, I think you should discard this pull request, and create a new one with only commit [77b6dfa](https://github.com/w3c/sdw/pull/920/commits/77b6dfa576dc81c8323592f7575efb0603cf7d34)

PR #921, opened 2017-06-13
Update to SSN-Usage document (only!)

arminhaller:
As of action https://www.w3.org/2015/spatial/track/actions/368
There is a revert in this of an old commit that has now been rejected. Please ignore.

Please accept/comment before the meeting!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 sdw
 ===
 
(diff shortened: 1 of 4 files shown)


maximelefrancois86:
fine with me as a first basis.

PR #924, opened 2017-06-13
implementation of ACTION-377 

maximelefrancois86:
Implemented ACTION-377 and added two new paragraphs 7.4 and 7.5 about to reflect discussion in https://www.w3.org/2015/spatial/wiki/Generic_vs_Specific_Modeling


dr-shorthair (commented):
Looks OK to me now.

PR #927, opened 2017-06-15
URLs for ISO and OGC references

dr-shorthair:
Updated to short/current form

### DIFF
--- a/ssn/config.js
+++ b/ssn/config.js
@@ -182,8 +182,7 @@ var respecConfig = {
        date: "July 2015"
       },
        "ISO-19109":{
-       href: "https://www.iso.org/standard/59193.html
-       ",
+       href: "https://www.iso.org/standard/59193.html",
        title: "Geographic information -- Rules for application schema",
        publisher: "ISO",
        date: "December 2015"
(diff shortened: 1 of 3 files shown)


ssn/config.js
```diff
@@ -169,26 +169,27 @@ var respecConfig = {
...
-       href: "http://www.iso.org/iso/catalogue_detail.htm?csnumber=57466",
+       href: "https://www.iso.org/standard/57466.html",
        title: "Geographic information -- Ontology -- Part 2: Rules for developing ontologies in the Web Ontology Language (OWL)",
        publisher: "ISO",
        date: "July 2015"
       },
        "ISO-19109":{
-       href: "http://www.iso.org/iso/catalogue_detail.htm?csnumber=59193",
+       href: "https://www.iso.org/standard/59193.html
+       ",
```
tidoust: This seems to have introduced a newline character, which makes the JSON invalid.

tidoust (changes requested):
See inline comment, newline character should be dropped. That looks good to me otherwise.

dr-shorthair:
OK - removed newline.

tidoust (approved):

PR #932, opened 2017-06-19
[QB4ST] Update markup of definitions and examples

tidoust:
This commit uses different syntax highlighting for definitions of the QB4ST Ontology and for code examples to be able to distinguish between the two.

All Turtle definitions have been updated to follow the same pattern, in particular to have one space before the final ";" or "." consistently across all code sections.

Also, the update disables the automatic syntax highlighter of ReSpec because it does not contain a Turtle highlighter and applies incorrect and inconsistent highlights to such code as a result.


BillSwirrl (approved):
Hi Francois - sorry for the slow response on this. I had missed that you had submitted this PR. I have now had a look and all seems fine to me - happy for you to merge

PR #934, opened 2017-06-19
Minor changes to the usage update document

rgcmme:


rgcmme:
I have added the usage information from http://pid.geoscience.gov.au/.
@arminhaller in some parts of the table I don't have a clear idea of the color scheme that you used. For me red=0, yellow=1 and green>1; but in your first coloring of rows in some cases this does not hold (maybe due to entailments and the top-level comments.
We should document the approach followed.

arminhaller:
Yes, the color changes due to entailment. But there may be errors.

rgcmme:
OK, can you then review the colors? I'm not clear about the reasoning followed for the entailment cases.

PR #954, opened 2017-08-06
Added implementation evidence from Alexandra Kokkinaki and Sergio Mendez and Krzysztof

arminhaller:


maximelefrancois86:
Looks good to me, but there are two Reference sections A and B, is this normal ?

PR #970, opened 2017-09-25
[BP] Adjust changes to changes section and SOTD

tidoust:
Added a "changes since 11 May 2017" section and adjusted the Status Of This Document section to note that this is a final release of the document.

Should more updates be made to the OGC section?

### DIFF
--- a/bp/index.html
+++ b/bp/index.html
@@ -447,9 +447,7 @@ $(document).ready( function() {
         datasets, <a>spatial data</a> is an especially useful addition to the Web of data.</p>
     </section>
     <section id="sotd">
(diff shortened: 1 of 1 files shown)


lvdbrink:
To my knowledge the vote for publishing this as an OGC Best Practice has already taken place and resulted in a yes. So we could just write it will be published as an OGC Best Practice.

tidoust:
Thanks @lvdbrink. I fixed the typo. I'm still not entirely clear what I should do in the SOTD for OGC. See [message I just sent to Scott](https://lists.w3.org/Archives/Public/public-sdw-wg/2017Sep/0029.html).

tidoust:
@lvdbrink, I adjusted the OGC part, using Scott's suggested text. Should be good to merge now.

PR #978, opened 2017-11-06
added implementation example of server specifications with apache server

maximelefrancois86:


ssn/integrated/proposed-server-specification.md
```diff
@@ -528,3 +528,119 @@ When operating a GET at http://www.w3.org/ns/ssn/systems/xyz with a Accept heade
...
+RewriteRule  ^(.*)/?$  /ns/sosa/oboe.ttl [env=redirect:1]
+
+RewriteCond  %{REQUEST_URI}  /ns/sosa/oboe$
+RewriteCond %{HTTP_ACCEPT} application/rdf\+xml
+RewriteRule  ^(.*)/?$  /ns/sosa/oboe.rdf [env=redirect:1]
+
+RewriteCond  %{REQUEST_URI}  /ns/sosa/oboe$
+RewriteCond %{HTTP_ACCEPT} text/html
+RewriteRule  ^(.*)/?$  https://www.w3.org/TR/vocab-ssn/#OBOE_Alignment [R=303,NE,env=redirect:1]
+
```
tidoust: The rewrite rules depend on the contents of the `Accept` HTTP header, so `Accept` should be added to the `Vary` HTTP response header. Apache does not do that automatically.

In other words, I believe the following directive should be added:

```apache
Header merge Vary "Accept"
```

ssn/integrated/proposed-server-specification.md
```diff
@@ -528,3 +528,119 @@ When operating a GET at http://www.w3.org/ns/ssn/systems/xyz with a Accept heade
...
+RewriteEngine On
+RewriteBase  /ns/sosa/
+
+
+# Documents
+
+<Files ~ "sosa\.rdf$">
+    Header set Content-Type "application/rdf+xml"
+    Header set Content-Disposition "filename= sosa.rdf"
```
tidoust: According to the [grammar in the RFC spec that defines Content-Disposition](https://tools.ietf.org/html/rfc6266#section-4.1), the value should include the disposition type:

```apache
Header set Content-Disposition "inline; filename=sosa.rdf"
```

I'm not exactly clear as to which disposition type we want to use: `inline` or `attachment`. I guess I would prefer `inline` which in theory allows the user to view the document in his browser (as opposed to saving it directly). But then, if the goal was actually to force the browser to save the document, `attachment` is the value we want here. I don't think all browsers support the disposition type in any case.

tidoust (commented):
See inline comments.

On top of these comments, I just wanted to record here one aspect that we already exchanged via email: there is no proper way to support the Accept HTTP header with rewrite directives in `.htaccess` files in Apache because of the `q` parameter. But that should only happen in corner cases that we can probably ignore here.

ssn/integrated/proposed-server-specification.md
maximelefrancois86: Let's go with `inline`, to me it's better user experience to get the document directly. 

A quick 2013 search led me to that 2013 report http://www.jtricks.com/bits/content_disposition.html that lists support for that header in common browsers -> apparently all ok

ssn/integrated/proposed-server-specification.md
maximelefrancois86: Agreed, `Vary` should be added but not on the "representations" URLs:

- Vary at http://www.w3.org/ns/sosa
- Vary at http://www.w3.org/ns/sosa/Actuation
- Vary at http://www.w3.org/ns/sosa/oboe
- no Vary at http://www.w3.org/ns/sosa/sosa.ttl
- no Vary at http://www.w3.org/ns/sosa/oboe.rdf

So I think this should do the job:

```
Header always set Vary "Accept" env=ttl
Header always set Vary "Accept" env=rdf
Header always set Vary "Accept" env=html
```

tidoust: Right. Note I will use `merge` and not `set` in my case, not to destroy other variation rules applied upper in the hierarchy. In particular, our server returns a `Vary: upgrade-insecure-requests` HTTP header that needs to stay.

ssn/integrated/proposed-server-specification.md
tidoust: Good, let's go for `inline`.

FWIW, "support" was not the right term here. What I meant was that the browser may not render the document even though `inline` is used, simply because it does not support the MIME type. That typically happens with Edge, which offers to save the Turtle version no matter what. But that's indeed orthogonal to supporting the `Content-Disposition` directive!

maximelefrancois86:
After some work from @tidoust and myself, the W3C server now implements these specifications (slightly updated), with content negotiation.

PR #986, opened 2017-12-11
Typo in ns prefix

dr-shorthair:
ssn-systems: --> ssn-system:


tidoust (changes requested):
Something must be missing, @dr-shorthair.

The only change in this Pull Request is the switch to transparent backgrounds on PNG images. That seems good, but not related to the typo mentioned?

PR #1002, opened 2018-01-18
Update of OWL-Time ontology, extended with rdfs:label, rdfs:comment a…

marianofl1971:
…nd skos:definition in Spanish for each definition where they were formerly just in English.


dr-shorthair:
That’s helpful. 
I found an additional stray assertion:

:intervalEquals owl:propertyDisjointWith :intervalIn . 

This may be correct, but is a semantic modification to what was agreed so can’t be slipped through in this way. 

Also: \n\n linefeeds inside strings are not reliably valid RDF. Turtle supports multi-line text inside triple quotes. 

Finally – the rdfs:Description type assertions are redundant in the .ttl and .nt versions at least. 
I can see how they might have sneaked in because of moving the translations to a separate block, but they can be removed from the .ttl and .nt versions at least.

dr-shorthair (changes requested):
That’s helpful.
I found an additional stray assertion:

:intervalEquals owl:propertyDisjointWith :intervalIn .

This may be correct, but is a semantic modification to what was agreed so can’t be slipped through in this way.

Also: \n\n linefeeds inside strings are not reliably valid RDF. Turtle supports multi-line text inside triple quotes.

Finally – the rdfs:Description type assertions are redundant in the .ttl and .nt versions at least.
I can see how they might have sneaked in because of moving the translations to a separate block, but they can be removed from the .ttl and .nt versions at least.

dr-shorthair (approved):
Looks good now. Thanks for the translations.

chris-little (commented):
The Spanish translation is welcome.  The structure looks fine to me.

marianofl1971:
Thanks a lot!

Mariano

2018-02-20 8:22 GMT+01:00 Simon Cox <notifications@github.com>:

PR #1007, opened 2018-02-11
Remove explicit declaration of xsd:dateTimeStamp.

dr-shorthair:
Some details of the definition (the subclass of rdfs:Literal axiom) were confusing some processors (Protege)
Is defined in XSD datatypes specification anyway.
Discovered when attempting to register with Bioportal.

### DIFF
--- a/time/rdf/time.jsonld
+++ b/time/rdf/time.jsonld
@@ -297,11 +297,6 @@
     "allValuesFrom" : "xsd:decimal",
     "onProperty" : "http://www.w3.org/2006/time#years"
   }, {
-    "@id" : "xsd:dateTimeStamp",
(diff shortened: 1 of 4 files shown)


tidoust:
Shouldn't it be removed from the [JSON-LD](https://github.com/w3c/sdw/blob/94c5eba8f294ee45004c66d348b12cb9a89cf39f/time/rdf/time.jsonld#L300) and [RDF/XML](https://github.com/w3c/sdw/blob/94c5eba8f294ee45004c66d348b12cb9a89cf39f/time/rdf/time.rdf#L51) serializations as well, at least for consistency?

dr-shorthair:
Yes, Done.

Issue #1114, opened 2019-01-09
OGC Coordinate Reference System Standards Working Group comments on OWL-Time


ogcscotts:
The attached comments on Time Ontology in OWL (OWL-Time) were made in response to the Open Geospatial Consortium request prior to OGC adoption of the specification. I have been asked to repeat them to the W3C forum.
 
Issue:
Terminology surrounding (temporal) coordinate system and (temporal) coordinate reference system.
 
Background:
OGC Abstract Specification Topic 2 = ISO 19111, Referencing by coordinates, defines the terms 'coordinate system' (CS) and 'coordinate reference system' (CRS). The data model includes a high level construct in which a CRS is composed of a CS and a 'datum' which defines the relationship of that CS to an object (usually but not necessarily the Earth).
 
The latest revision of OGC Abstract Specification Topic 2 (awaiting formal publication but text available at OGC 18-005r3) = ISO 19111:2019 (also awaiting formal publication, expected January 2019), has added detail to the elements defining a temporal coordinate reference system[1]. The same structure is used for temporal CRS as for other subtypes of (spatial) CRS, that is the high level construct is a temporal coordinate reference system which is composed of a temporal coordinate system and a temporal datum. A Topic 2/19111 Coordinate Reference System maps to the concept in 19108 called Coordinate System. This is confusing because of the different uses of the term 'coordinate system'. But leaving that ambiguity aside, see OGC 18-005r3 Annex D.1 for an overview of Topic 2 discussion of Temporal CRS.
[1] Historical note. During the revision of Topic 2 that led to 19111:2007, this detail was understood to be added to a proposed future revision of ISO 19108:2002. ISO 19111:2007 included a temporal CRS class as a stub but with no detail, just a reference to 19108. That revision of 19108 has never happened. As a consequence the missing detail was added to the recent 19111/Topic 2 revision, and is being included in the ongoing CRS WKT revision (ISO 19162).
 
Suggestions:
Before being adopted as an OGC specification, the Owl-Time documentation should be updated to acknowledge OGC Abstract Specification Topic 2 and mention the concept of a Temporal CRS. The following edits are suggested:
 
i) In 3.2 (Temporal reference systems, clocks, calendars), paragraph 6 add reference to ISO 19111 (OGC Topic 2) temporal reference system: " ... using a representation of temporal position in a temporal coordinate system [iso19108] or temporal coordinate reference system [iso19111], [OGC Abstract specification Topic 2], i.e. ..."
 
[...]

PR #1116, opened 2019-03-06
OWL-Time - Editorial changes in response to #1114

dr-shorthair:
Make editorial changes requested by Roger Lott in the course of OGC adoption (documented in #1114 )

For preview see https://rawgit.com/w3c/sdw/time-issue1114-take2/time/


tidoust (approved):
This looks good to me! I pushed an update to fix a couple of typos in the errata entry and use dated URL links to reference the incriminated sections.

FWIW, through the "diff settings" menu on GitHub, one can "hide whitespace changes", but it seems better to restrict actual changes to a bare minimum so that we don't have to do that each time a pull request gets created against the specification, so thanks for fixing the whitespace normalization issue!

dr-shorthair:
So how do we now wait on the OGC vote before merging and then re-publishing? @lvdbrink

PR #1138, opened 2019-07-14
OWL-Time - add IANA considerations section

dr-shorthair:
for the link relations that have already been registered 

This transcribes the [information actually submitted to IANA for the registration of link relations for time intervals](https://github.com/link-relations/registry/issues/4) that happened December 2018. Requested by IANA commenters to close the loop. Should have been included originally, but has no effect on users of OWL-Time.


dr-shorthair:
See preview at https://rawgit.com/w3c/sdw/time-iana-considerations/time/#iana-links

chris-little:
@dr-shorthair . I have been trying to understand why there are 15 rather than the original 13 Allen relations. Is there a reason why para 4 (disjoint) and para 9 (in) are one statement rather than two separate ones, matching the style of all the others? I suppose it gives a kind of symmetry to the other 'edge case' para 6 (equality).
But 4 is covered by cases 1 and 15, and 9 by 5,6,7,8,14 and 15.

Has anyone done any work on the optimal variations of the orginal 13 relations,perhaps including the instant as a special interval?

(It is a bit late in the day and i'm not feeling very smart) Chris

chris-little:
@dr-shorthair , (Now had a G&T). The extra two relations are the two general ones that we added in the  OWL-Time. Perhaps we could order them as the original 13 Allen relations and then the two general ones, or vice versa, to make it easier for people to relate to the underlaying algebra.

dr-shorthair:
The reasoning is spelled out in the new document - see https://rawgit.com/w3c/sdw/time-entity-relations/proposals/time-entity-relations/index.html#motivation 

I would be OK with re-ordering the list, but this is the order that they appear in the IANA registry - https://www.iana.org/assignments/link-relations/link-relations.xhtml#link-relations-1

lvdbrink:
@dr-shorthair can you address the warnings ReSpec gives about normative references in informative sections? 

> Normative references in informative sections are not allowed. Remove '!' from the start of the reference [[!owl2-quick-reference]] at: 1.

Once you fix those I will merge!

dr-shorthair:
Done. 
It seems that ReSpec has been changed again so the prior assumption that if a section had no @class assignment then it was 'normative' no longer applies.

PR #1149, opened 2019-12-05
Purely editorial changes

PeterParslow:
Just a few editorial things whilst reading.
As a largely passive member, I am supportive of the idea of re-chartering as described. I remain confident that other OS colleagues will be more active than me!


lvdbrink:
Thank you Peter!

PR #1166, opened 2020-02-18
Updated to correct teleconference info

MichaelGordon:


lvdbrink:
When @tguild and I discussed the charter update it seemed like a good idea to us to go back to monthly telco's. I'll create an issue to discuss this further.

MichaelGordon:
Ah apologies, I thought it was just a copy and paste from the original. No probs 😊

PR #1207, opened 2020-10-21
Removing Common Concerns section and adding content to regulators

jabhay:
Integration authorities, etc., appeared to fit best with regulator
perspective. As such, the content has been weaved in.


jabhay:
Concept reviewed by the team. We're happy to merge it.

PR #1213, opened 2020-11-03
Updated Responsible Use graphics

rjksmith:
Replaced cartoons with icons
Replaced editor's notes with key question notes
Fixed Respec error


jabhay (approved):
Looks good!

PR #1219, opened 2020-11-18
Update index.html

edpars0ns:
Added some content to the intro, moved the spatial is special section up in the doc and added some text to the spatio-temporal section


jabhay (approved):
looks good

PR #1227, opened 2020-12-07
Responsible Use of Geo: Update user perspective

rjksmith:
Use of 'location' for public engagement in preference to 'spatial data'


jabhay (approved):
looks good to me.

PR #1228, opened 2020-12-10
This change makes all uses of data plural for consistency

jabhay:
We had varying plurality of the word data. This change makes
all of them plural to achieve consistency.


chris-little:
We approve! ;-)

PR #1242, opened 2021-01-21
Add OGC logo to config

lvdbrink:
We consider documents from the SDWIG to be joint from W3C and OGC and should therefore include the OGC logo. 

Sorry this was caught just too late...


tguild (approved):
Fine with me. I would rather wait for additional changes before republishing. At that point I will see if this type of document and stage (Working Draft for a Note) can auto-publish on subsequent updates. It should spend some time as a First Public Working Draft first though.

jabhay (approved):
Looks good

PR #1257, opened 2021-05-04
Update index.html

kjano:
A proposal for an updated text for hasMember.

### DIFF
--- a/ssn-extensions/index.html
+++ b/ssn-extensions/index.html
@@ -386,7 +386,7 @@ See <a href="#container-property-rule">Inheriting observation properties from a 
   </tr>
 <tr>
 <td>Definition:</td>
-<td>Link to a member of a collection of observations that share the same value for one or more of the characteristic properties</td>
+<td>Links an Observation or ObservationCollection that shares the same value for one or more of the characteristic properties to an ObservationCollection.</td>
 </tr>
 <tr>
 <td>Instance of:</td>


dr-shorthair (approved):

kjano:
I saw that you approved the change. Thanks a lot. I do not want to mess with your workflow; should  I merge or will you or Linda?

dr-shorthair:
Process has been a bit loose in these byways. 
Would be good for @lvdbrink or @jabhay to be aware of this work, so let's ask them. 

Would be good to fold ssn-ext into vocab-ssn, but I don't think we can do that without chartering a new W3C working group. 

Meanwhile, CSIRO has let its membership of W3C lapse, after ca. 25 years. I was responsible for the organization joining in 1996 or thereabouts, but it has been with the computer scientists for the last 20, and they have lost interest. So I will have to manage my status to be able to continue to contribute ...

kjano:
@dr-shorthair Oh! I had no idea about the W3C issue :-(. Should I sync these changes we made with the RDF file, so far they are only in the documentation, correct?

dr-shorthair:
I think it is OK in the files now - https://github.com/w3c/sdw/blob/gh-pages/ssn/integrated/examples/tree-height.ttl#L57 

I did get bamboozled by Francois' HTML tricks though.

lvdbrink:
Perhaps @tidoust could help us out here...

PR #1302, opened 2021-12-20
Update SDWBP FAIR thoughts.asciidoc

jvanulde:
Just a suggestion to include verification.


PeterParslow:
I think that's a sensible suggestion for somewhere in the best practice. Not sure it sits comfortably with this bit though, which is there to relate our work to FAIR.

The Q and "+" challenges have been made publicly; have I so far missed the suggestion to add V?

PeterParslow:
2022-02-24: meeting agreed that it is good to include something about verification somewhere in the Best Practice, but that it is not particularly related to FAIR.

@jvanulde Please raise an issue on the BP as a whole.

Issue #1313, opened 2022-01-12
Best practice 3, 2nd note - wrong word "prescriptive"


PeterParslow:
The second note of Best Practice 3 contains the sentence

"Of course, often target resources and the links that refer to them are maintained by different parties, so such hints should be assumed as prescriptive; they may or may not turn out to be true."

(line 1621 of bp/index.html)

I think the word "prescriptive" is wrong, or at least being used in a way which I'm not familiar with. Normally, describing something as "prescriptive" means it is to do with the enforcement of a rule, either legally or by long established tradition. The idea being described here (that the information may not actually be true) is nearer to what would be meant by "descriptive" or perhaps "indicative".

It may be better to rephrase the whole sentence to something like:

"Of course, often target resources are maintained by different parties, so information provided with the links that refer to them may or may not turn out to be true when the link is traversed."

PeterParslow:
Fixed by https://github.com/w3c/sdw/pull/1322

PR #1322, opened 2022-01-13
wrong use of "prescriptive"

PeterParslow:
Fix as described in https://github.com/w3c/sdw/issues/1313


situx (approved):
Thank you for spotting! Merging it in now

PR #1323, opened 2022-01-19
Chris little coverages patch 1

chris-little:
@lvdbrink @6a6d74 Hopefully I have improved the coverage section and a few other paragraphs. I have also altered some instances of the word `coverage` where it has been used to mean `extent`. Towards the end of the document, a couple of occurences of the word `coverage` are ambiguous, and I have left these for discussion. 

I have not found or inserted any images.


situx (approved):
I like the changes to the text, but I have two scope questions:
- Should 3D coverages be discussed?
- For raster data, there is raster algebra used to transform raster representations. Is that worth mentioning?

I think some images of the several coverage types would be really helpful to give the reader an idea. Maybe that would be a separate issue?

chris-little:
> I like the changes to the text, but I have two scope questions:
> >     * Should 3D coverages be discussed?
>     * For raster data, there is raster algebra used to transform raster representations. Is that worth mentioning?
> I think some images of the several coverage types would be really helpful to give the reader an idea. Maybe that would be a separate issue?

@situx I think images in a separate issue is a good idea.

I had assumed that the text was largely nD agnostic, except for the 1D timeseries case, which many do not think of as a coverage. 

Strictly, a 3D coverage (assuming you mean basically (x,y,z) rather than (x,y,t) for example) should have a single, 3D coordinate system, whereas many use cases have a separate third CRS, and the 'coverage' is actually a 'stack' of coverages or a time series of coverages. I was trying to avoid mentioning this detail.

Raster algebra is a 2D, gridded coverage special case of geospatial algebra. Mentioning it would reinforce the perception that coverages are only 2D rasters images. Perhaps we need a section on geospatial algebraic processing and also the topological intersection algebra (DE-9IM) of Egenhofer and Herring?

situx:
Great, I will create another issue for adding coverage images as for the geospatial algebra I will add this point to the next meeting to discuss the Spatial Data On The Web Best Practices.

lvdbrink (approved):

situx:
Since we have two approvals and another issue has been created for the graphics I am merging this one in

PR #1332, opened 2022-01-27
Reduce ambiguity

PeterParslow:
The list is quite a long way above (with several bullet points between), so including the phrase "Bespoke API" helps.

"as" is ambiguous (I initially read it as about to introduce the name/description of the third option); "because" is less so.

### DIFF
--- a/bp/index.html
+++ b/bp/index.html
@@ -3459,7 +3459,7 @@ Content-type: application/geo+json
             href="https://www.w3.org/TR/dwbp/#APIHttpVerbs">Best Practice 24: Use Web Standards 
           as the foundation of APIs</a>, but include APIs targeted at a specific programming 
           language, for example, JavaScript.</p>
-        <p>In the list of options above, a third option is included as sharing <a>spatial data</a>
+        <p>In the list of options above, the third option - Bespoke API - is included because sharing <a>spatial data</a>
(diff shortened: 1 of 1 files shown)


PeterParslow:
I hope someone is re-writing much of this section to include description of the more contemporary OGC APIs

PeterParslow:
And what's the relationship between "Bespoke APIs" and "Convenience APIs"?

chris-little:
> And what's the relationship between "Bespoke APIs" and "Convenience APIs"?

@PeterParslow Complexity vs Simplicity?

cportele:
> And what's the relationship between "Bespoke APIs" and "Convenience APIs"?

A "bespoke API" is an API that largely ignores standardized API building blocks (using OGC API terminology).

A "convenience API" is an API that is designed for a specific usage (convenient for clients supporting that use case).

There is no strict relationship. Sometimes an API has to be "bespoke" to be "convenient", but for other uses an OGC Web API (not "bespoke") can also be "convenient". BP 12 basically says that you should look at the user needs in your API design.

When we wrote BP 12 five years ago the point was that WFS etc are for most web uses/users not easy or convenient to use. In that sense the text implies that often convenience implies bespoke, but there has been some progress and the whole discussion needs a review.

Issue #1292, opened 2021-11-22
[SDW Best Practices Update]: Review updates for Gaps in current practices


situx:
Review BP for any updates necessary to the Gaps in current practice section and any things from the Gaps section that have matured enough to feed into the main BP. I already did some work on this. See #1259

Issue #1259, opened 2021-05-05
Analysis of Gaps in current practice


lvdbrink:
In the BP subgroup meeting of 20 April we said we wanted to analyse the [Gaps](https://www.w3.org/TR/sdw-bp/#conclusions) section in the Best Practices document to see if now, after almost 4 years they were still gaps or perhaps a good practice has since emerged and some of them can be promoted to best practices. 

**Addition:** And also to see if there are _new_ gaps, i.e. things related to geo+web that people are struggling with and missing a best practice.

lvdbrink:
My first quick analysis

**[13.1] identifies gaps concerning the sharing of geometries on the Web**. It describes 3 concerns: 
1. We could not find one single way to express geometries that is recommendable in all cases. To my knowledge this is still the case, but I'm not sure I consider that a 'gap'. There are simply different recommendable ways of doing it for different use cases. We describe that in BP4. We might be able to improve on the text in that BP based on new insights. 
2. We couldn't identify a best practice for publishing geometries of different precision and accuracy. I think that is still the case. I don't know of new practice that adresses this. 
3. We couldn't identify a best practice for publishing and accessing geometries in different CRS. Here, I see a practice emerging. There are new developments, such as a standardized way to deal with this in OGC API Features (already published standard), GeoSPARQL 1.1 (on the way). 
   We described a possible way forward with CRS conneg, but I haven't seen that gaining traction. It is only implemented in the Netherlands. Perhaps we should consider removing it. 

**[13.2] identifies the lack of a go-to spatial data vocabulary**. GeoSPARQL is mentioned there; my first thought is that GeoSPARQL 1.1, which is being developed now, might fill that gap. Its vocabulary is richer than it was in 1.0 and includes some useful properties like centroid and bounding box, to mention a few. 
Of course, it's too early to proclaim this a best practice on the basis of GeoSPARQL 1.1, because it will not be a standard for a while yet, let alone an implemented one for which we can find good practice in the wild. 

**[13.3] identifies a gap in standardized dataset and service descriptions for data portals**. It mentions the OGC working on such a standard, but to my knowledge no such work is progressing at the moment. Maybe we were thinking of GeoDCAT-AP, a European standard; there were plans for picking that up within OGC but I think it has stalled. 
Meanwhile, there is a new version of GeoDCAT-AP based on DCAT 2.0. Even though it is not a worldwide standard, we might say, well there is a good practice we see in the wild in Europe, let's promote this to a BP. That's one way forward. 

[...]

PR #1334, opened 2022-02-09
Lvdb gaps

lvdbrink:
This PR addresses part of the issues coming out of my analysis of the Gaps section (#1292, #1259). It only focusses on paragraph 13.1 of the Gaps section.

- Removed first paragraph of 13.1 about the question of different serialization formats to choose from
- Removed 13.1 paragraph about CRS conneg
- Addition to BP8: a fifth way to provide CRS information to the user (using OGC API Features part 2)
- Addition of new BP Support requesting and returning geometries in a specific requested CRS

See #1259 for argumentation.


situx:
A general question when reviewing this:

Are we using the term CRS or SRS, and is there a valid reason for that if we are using CRS?

You are still missing one GeoSPARQL example which could look like the one below which filters by CRS84:

SELECT ?geo_wkt WHERE {
  ?geo geo:asWKT ?geo_wkt .
  FILTER(geof:getSRID(?geo_wkt)==<http://www.opengis.net/def/crs/OGC/1.3/CRS84>)
}

Apart from that I think the text is good!

situx (changes requested):
GeoSPARQL example is still missing and a question concerning the use of the term SRS and how to deal with e.g. Geocoding as something that does not indicate a CRS could be helpful

andrea-perego:
> You are still missing one GeoSPARQL example which could look like the one below which filters by CRS84:
> 
> SELECT ?geo_wkt WHERE {
> ?geo geo:asWKT ?geo_wkt .
> FILTER(geof:getSRID(?geo_wkt)==http://www.opengis.net/def/crs/OGC/1.3/CRS84)
> }

About the example, I wonder whether `geof:getSRID` returns the SRS also when it is not explicitly specified in the geometry literal (excuse my ignorance).

Shouldn't this be the case, it would be probably better to revise the example to use either another SRS or another geometry literal encoding.

situx:
In GeoSPARQL 1.0 the situation is as follows:
- WKT Literals without a CRS definition are assumed to use the CRS84 CRS
- GML Literals without a CRS definition are allowed but would probably fail the test of a GML parser (not sure though)

In GeoSPARQL 1.1 the situation will be as follows:
- GeoJSON literals and KML literals have a fixed CRS by definition of the literal format
- DGGS literals without a CRS definition are to my knowledge not defined correctly, so they should not parse correctly

The specification (GeoSPARQL 1.0 and 1.1) only says that the CRS of the geometry literal is to be returned. How to treat errors is up to the implementers to a certain extent.
I believe that the triple store would return a general error message here.

lvdbrink:
Added the GeoSPARQL example provided by @situx above. 

> Are we using the term CRS or SRS, and is there a valid reason for that if we are using CRS?

Currently the test is predominantly using the term CRS and I think appropriately, because we are talking about coordinate reference systems. In the context of requesting data in a specific reference system (which this PR is about), CRS is the correct term as we mean to talk about coordinate reference systems. However, I can see the benefit in talking about spatial reference systems in the larger context of the BP document. But this would be a separate issue.

BillSwirrl (commented):
Hi Linda, your updates look good to me.  I only have some minor comments and suggestions.

bpconfig.js line 255.  "Clements" should be "Clemens"

New example on Content-crs HTTP header in BP8 looks good

New best practice "Support requesting and returning geometries in a specific CRS":

Line 3787

I would suggest being less detailed on the mechanics of re-projection, because there could be lots of ways to do it.  Suggested alternative (not perfect either, so feel free to use, edit or ignore as you like!)

 <p>Offering geospatial data in different <a>CRS</a> in practice means the data needs to be transformed from the original <a>CRS</a> to other supported <a>CRS</a>s. Database software and libraries for most programming languages are available which can do this.  Reprojection can be done in advance and stored, or calculated at the time of a request. If reprojecting on request, since geographic data can be large and complex, it is recommended to cache the converted data to eliminate the need of reprojecting the same data more than once. </p>
 
 The example that goes with this is really useful.
 
 And I agree the updates to  the 'different representations of geometries' section is an improvement.

lvdbrink:
@situx is it okay to merge given my reply to your review?

situx (approved):

situx:
Looks good to me Linda. Should be ready to merge

Issue #1179, opened 2020-06-26
Updates to the Spatial Data on the Web Best Practice NOTE


6a6d74:
Discussion at the teleconf (Thu 25 Jun 2020) concluded that we should recharter as a Working Group scoped to update the Best Practice NOTE (and potentially other stuff).

Clara Boyd (Ordnance Survey) supported by the IG as the new lead editor.

ACTION: Jeremy, Linda, and Clara (with support from Ted) to begin working on a new WG Charter.

My notes from the teleconf below. Official minutes [here](https://www.w3.org/2020/06/25-sdw-minutes.html).

——
(Ed & RobS)
Discussion about ethics of location sharing - “here be dragons”
(Clemens)
General conclusion: structure and content mostly good
… only needs updates
… especially examples (some don’t work any more | HTTP 404)
… plus new examples
… additional topics:
… OGC API work - update the API chapter and “what’s wrong with SDIs”
… there’s a very heavy RDF bias in the Relationships section; mostly the OGC APIs use “web linking”
… increasing use of “tile” concepts (noting the work in OGC to standardise the conceptual model and the tiles API) - there is practice in this space, even if not yet standardised
… dynamic CRS etc. (relating to BP8)
(Bill)
… DCAT version 2 (and potential to feed into ongoing DCAT discussions)
… schema.org updates - there’s been lots of implementation in this space
… linking spatial and non-spatial data - this is much easier with OGC API
… JSON LD 1.1 makes it much easier to use with GeoJSON - this is a prime candidate for how examples can be updated
(RobA)
… FAIR Principles
… more than just DCAT; people are sharing data in ad-hoc ways (e.g. PDF docs); there’s lots of work to improve with ad-hoc data sharing [to improve (re-)usability of that data] using profiles and JSON-LD
… most people do JSON-LD poorly thanks to the tooling; the @context files are often huge
… best practice around modularity of @context files (separation of concerns), registering profiles (for re-use), and linked data descriptions of profiles

akuckartz:
> … there’s a very heavy RDF bias in the Relationships section; mostly the OGC APIs use “web linking”

The RDF "bias" hopefully was considered to be a good thing.

rob-metalinkage:
note: my comment was that data models (as metadata) are being shared in ad-hoc ways. I wasnt attempting to cover the more general issue of data sharing ;-)

rob-metalinkage:
[...]

PR #1336, opened 2022-02-14
Add reference to Fair into Reality report

KoalaGeo:
https://ec.europa.eu/info/sites/default/files/turning_fair_into_reality_0.pdf

### DIFF
--- a/bp/bpconfig.js
+++ b/bp/bpconfig.js
@@ -258,6 +258,12 @@ var respecConfig = {
 	"title":"Turning FAIR into reality. Final report and action plan from the European Commission expert group on FAIR data",
 	"date": "2018"
       },
+      "FAIR-DMM":{
+        "href":"https://doi.org/10.15497/rda00050",
+        "publisher": "Research Data Alliance",
+	"title":"FAIR Data Maturity Model. Specification and Guidelines (1.0)",
+        "date": "2020"
+      },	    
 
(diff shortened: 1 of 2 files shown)


andrea-perego:
If I may, I would suggest referring to this document via its DOI, to avoid the broken link issue in the future:

https://doi.org/10.2777/1524

Also, it would be probably better to add it as a bibliographic reference in `bpconfig.js`:

````javascript
"FAIR-XG":{
  href:"https://doi.org/10.2777/1524",
  title:"Turning FAIR into reality. Final report and action plan from the European Commission expert group on FAIR data",
  publisher: "Publications Office",
  date: "2018"
},
````

BTW, another possibly useful reference is the one I mentioned in https://github.com/w3c/sdw/issues/1179#issuecomment-951599100 , which is the result of work that was triggered by the report above, and tries to define precise criteria to assess compliance with the FAIR principles.

````javascript
"FAIR-DMM":{
  href:"https://doi.org/10.15497/rda00050",
  title:"FAIR Data Maturity Model. Specification and Guidelines (1.0)",
  publisher: "Research Data Alliance",
  date: "2020"
},
````

KoalaGeo:
@andrea-perego amended as suggested

andrea-perego:
> @andrea-perego amended as suggested

Thanks!

PeterParslow (approved):

PR #1344, opened 2022-03-30
BP 11 life-cycle rules

### LINKED ISSUE #1329: BP 11 UK Digital National Framework disappeared years ago
Best Practice 11 says:
"For example, the UK's Digital National Framework policy states that data publishers must provide these lifecycle rules."

The UK's "DNF" strategy disappeared some years ago, and the current UK location strategy is nowhere near as detailed. Hopefully, someone can think of another example supporting

"A description of the lifecycle of the Spatial Things (e.g. what triggers a change and whether those changes are versioned etc.) should also be provided"

INSPIRE ISDSS regulation and "Life-cycle of Spatial Objects" and D2.5 clause 9.7 "Spatial object life-cycle" are about how to represent the information, not a simple statement that you should.

PeterParslow:
Swap UK DNF reference for INSPIRE one


andrea-perego:
@PeterParslow , may I suggest using the corresponding [ELI URI](https://eur-lex.europa.eu/eli-register/about.html) to link to the INSPIRE Regulation?

http://data.europa.eu/eli/reg/2010/1089/oj

PR #1355, opened 2022-05-13
Update to Gaps section

### LINKED ISSUE #1292: [SDW Best Practices Update]: Review updates for Gaps in current practices

lvdbrink:
Includes updates to all Gaps sections as needed, except for 13.1 which was already addressed in #1334 

- [x] 13.2 Spatial Vocabularies: added mention of GeoSPARQL update work, including mention of CRS ontology, and short description of additions to GeoSPARQL 1.1 vocabulary.
- [x] [13.3 Describing dataset structure and service behaviors
- [ ] 13.4 Publishing dynamic and large datasets on the Web
- [ ] 13.5 Helping software understand units of measure
- [x] 13.6 Defining that two places are the same
- [ ] 13.7 Discovering what refers to a Spatial Thing

When finished, closes #1292


lvdbrink:
13.4 Publishing dynamic and large datasets on the Web: PROPOSAL to leave this section as-is unless a case is made to convert this into a BP e.g. around OGC EDR. 

13.5 Helping software understand units of measure: PROPOSAL to leave this section as-is unless examples are presented to us of providing UoM info in the way described in this section. 

13.7 Discovering what refers to a Spatial Thing: PROPOSAL to leave this section as-is unless examples are presented to us  of using VOID in the way described in this section.

situx:
I think your changes are good, but I wonder if other spatial types should be mentioned as well. For example coverages, 3D meshes, and building information management data are all spatial data. Should they be mentioned along with geospatial data in the gaps section?

PeterParslow:
"other spatial types should be mentioned as well. For example coverages, 3D meshes, and building information management data are all spatial data" - agree in principle.

But, BIM is more of an "application schema" that uses spatial types. It does use a different subset of the "vector" spatial representations to that usually used in GIS, but it's not unique in that.

- GIS generally uses "surveyed" data, where curves are represented by linear interpolation (in the relevant CRS) between points along the curve.
- BIM (and some legal GIS such as maritime boundaries) prefers parametric curves, defined by mathematical arcs.

This reflects the origin of BIM & much BIM data in design drawings, rather than 'as built' surveys. 

So if we want to go that detail, we would divide "vector" spatial types into "surveyed" and "modelled" (or some such - those are terms I just created here). I suspect coverage aficionados would claim that 3D meshes are a kind of coverage. It all depends on what aspect one uses to partition the "spatial type" conceptual space!

lvdbrink:
If we want to mention other spatial data types besides geospatial data, shouldn't that be an issue for the entire document, not just the Gaps section? 

Happy to briefly mention other spatial data, but I don't have enough expertise to go into detail there.

situx:
I do not think we need to go into that much detail. However, I know that there are gaps in standardization, at least for ontologies, but also for groups working on these specific issues. In any case, it is the gaps section.
I also do not think we need to widen the scope of the document. Coverages are discussed in any case.

PR #1376, opened 2022-11-07
Replace example format of Shapefile with GeoPackage

KoalaGeo:
Use format which is an open standard in preference to a proprietary format


situx:
Thanks, this makes sense to me.
Even though shapefiles are still commonly used (to my dismay), we should set a better example in this document.

situx (approved):

PR #1382, opened 2022-12-21
Update editors & contributors

### LINKED ISSUE #1350: Editors/Contributors will need updating
Probably to be completed at the end.

PeterParslow:
Added Timo as an editors in this version, and Iza as a contributor.

See #1350

### DIFF
--- a/bp/bpconfig.js
+++ b/bp/bpconfig.js
@@ -21,7 +21,12 @@ var respecConfig = {
 			name: "Payam Barnaghi",
 			company: "University of Surrey",
 			companyURL: "https://www.surrey.ac.uk/"
-		}],
+    }, {
+      name: "Timo Homburg",
+      company: "Hochschule Mainz University Of Applied Sciences",
+      companyURL: "https://www.hs-mainz.de/"
+    }, 
+    ],
     otherLinks: [
     {
       key: "Contributors",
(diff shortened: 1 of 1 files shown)


situx (approved):
Good from my side. I added my ORCID, as well, and I wonder if we should do that for more people. Is that a common thing to do?
I see that on some W3C specifications like https://www.w3.org/TR/annotation-model/

lvdbrink:
> Good from my side. I added my ORCID, as well, and I wonder if we should do that for more people. Is that a common thing to do? I see that on some W3C specifications like https://www.w3.org/TR/annotation-model/

I don't have one. 

Something to consider: should Jeremy, Payam and I still be editors in the new edition? I feel like I could also be considered a contributor on this one. Respec has `formerEditors`, which we could use to list the editors of the first edition.

PeterParslow:
I wasn't sure whether to interpret editors & contributors as "of this edition"; I went for "of the whole thing" - so I didn't remove any.
It may make sense to distinguish the editors, but leave the contributors in one list (I doubt we have entirely removed anyone's earlier contribution!).

That said, I wasn't sure what the distinction is in W3C between editor & contributor. I took a view that Linda & Timo have carried the bulk of the organising work so should be listed as editors (in the sense of "project leaders" really!)

situx:
I would agree with Peter here. My impression has been that the list of editors and contributors would be comprised of all people who contributed to the original document and the update.
But I also have no guideline of how that is usually handled.
We could use former editors, but a former contributors field would also come in handy. I cannot find that in Respec, however.

lvdbrink:
Let's keep it as it is now then - it was just a thought. I don't know of guidelines about this, either.

PR #1383, opened 2022-12-22
Remove Editors note about missing namespace

PeterParslow:
Because someone has already added them, so the note was no longer true.


situx (approved):
Thank you for spotting this. I will merge it in right away.

Issue #1084, opened 2018-10-24
Evaluate DCAT changes in terms of impacts to best practices


MichaelGordon:
At the Lyon face to face meeting, it was highlighted that the [Dataset Exchange Working Group](https://www.w3.org/2017/dxwg/wiki/Main_Page) are working on a new draft of Data Catalog Vocabulary (DCAT) - https://www.w3.org/TR/vocab-dcat-2 

As the Spatial Data on the Web Best Practices advises use of DCAT in various places (for example https://www.w3.org/TR/dwbp/#DescriptiveMetadata) the changes to DCAT should be evaluated in terms of the potential impacts to best practices and feedback provided.

MichaelGordon:
Comments from @cportele to public DXWG mailing list:

"I want to express concerns about the direction of DCAT with respect to distributions and services. The rationale for making services first class citizens in data catalogs is not clear to me. The model of DCAT 1.0 is in this area, in my view, clearer and more useful from an end user perspective. The items registered in a *data* catalog should be restricted to data (that is: datasets - unless you register individual items). As a user I go to a data catalog looking for datasets, not services; once I have found some candidate datasets then I want to understand how to access the data, whether that is a file download, through webpages or an API. This relationship is described in the Data on the Web Best Practices (https://www.w3.org/TR/dwbp/#context), too. 

DCAT 1.0 does not provide sufficient detail or guidance for documenting the different kinds of distributions. A downloadable file is straightforward, but an API could benefit from more information than what DCAT 1.0 supports. From the UCR document I expected that the reversion of DCAT would improve Distribution instead of adding Services as additional, separate resources in the mix. The separation between Distribution and Service looks blurred to me, with overlap and redundancies in their properties. Please reconsider this design decision.

In my view, the current direction is repeating some of the mistakes that the spatial data community made in their standards.

I have a few additional questions that are related to this and also related to the topic of profiles:

[...]

PR #1384, opened 2022-12-22
Update index.html - DCAT2

### LINKED ISSUE #1297: [SDW Best Practices Update]: DCAT2 Update
DCAT 2 update

PeterParslow:
Changes related to https://github.com/w3c/sdw/issues/1297

Some DCAT -> DCAT2 changes had already been made

Should we remove DCAT1 from the informative references? If so, should we then re-use the id/link VOCAB-DCAT to mean DCAT2?

### DIFF
--- a/bp/index.html
+++ b/bp/index.html
@@ -758,7 +758,7 @@ function bpCrossRefs() {
           <tr>
             <td><code>dcat</code></td>
             <td>http://www.w3.org/ns/dcat#</td>
-            <td>Data Catalog Vocabulary (DCAT) [[VOCAB-DCAT]]</td>
+            <td>Data Catalog Vocabulary (DCAT) [[VOCAB-DCAT-2]]</td>
           </tr>
           <tr>
             <td><code>dcterms</code></td>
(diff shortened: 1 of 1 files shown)


situx:
The changes look good.
I would be in favor of removing the DCAT1 reference. After all, DCAT-2 has been around for almost three years now.

PeterParslow:
@situx : there are two ways to remove the DCAT 1 reference:

- just remove it from the bibliography. leaving the VOCAB-DCAT-2 references as they are
- switch all the DCAT references to use VOCAB-DCAT but make that point to v2

Which do you think works best?

situx:
I would go with the first option. Then it is immediately clear by just looking at the reference in the text that we point to version 2 of DCAT

PeterParslow:
Now I discover that I can't see how to do that - I can't see where the "informative references" entries come from, other than those in bpconfig.js "localBiblio" array (which doesn't contain either VOCAB-DCAT entry).

index.html has a comment referring to "SPECREF" and a commented out call to 

//    preProcess: [bpCrossRefs]

Does anyone know how the informative references section is assembled?

situx:
I think it is in the bpCrossRefs file, and an additional set of standard references was provided during the compilation of Respec. DCAT is likely common enough to be among them.
Also at the end of the document there is a call to bpCrossRefs()
Does it not work if you add DCAT-2 to the localBiblio?

PeterParslow:
"I think it is in the bpCrossRefs file" - that was my though from the preProcess statement. But I can't see any bpCrossRefs file in the repository.
I didn't need to add VOCAB-DCAT-2, someone had already done that. The remaining bit is to remove VOCAB-DCAT because there are no more references to it.

PeterParslow:
I was also wondering how to change what is included about issue 1084. What gets pulled in is the initial description of that issue. It would be fairer now to say something like:

"The Best Practices have been updated to use DCAT2, but not to check if DCAT2 itself introduces any new best practices beyond those already described."

I have put that as a comment on the issue, but the summary that is imported to the BP document still implies it hasn't been done - the reader would need to click through.

As an alternative, perhaps we close issue 1084? Noone in the group has proposed a new best practice, which suggests either that DCAT2 does not offer any new best practices, or that the new things it does offer are not yet widely used.

situx:
I think it makes sense to close issue #1084 
As for the DCAT issue. Could we not update all references to DCAT-2 and open a new issue for removing DCAT from the bibliography?
As I also do not see how to do that currently, it probably needs further investigation

PeterParslow:
All the references are now to DCAT-2. The new issue is #1387

lvdbrink (approved):
Looks good to me. 

Not sure if you figured out that a lot of bib references are not defined locally but in specref, see https://www.specref.org/?q=dcat

PeterParslow:
"Not sure if you figured out that a lot of bib references are not defined locally but in specref, see https://www.specref.org/?q=dcat"

I hadn't stumbled across that bit of "hidden knowledge" :)

situx (approved):

Issue #1146, opened 2019-11-19
Should the best practices mention ethical use of geo, and if so how?


MichaelGordon:
Raised during discussion of WebVMT at OGC Toulouse TC with @rjksmith 

WebVMT files provide map presentation and annotation synchronised to video content, including animation support, and more generally any form of geolocation data that is time-aligned with audio or video content.

This creates a potential issue of facial recognition combined with geospatial information knowing exactly who and where someone is. Therefore should the best practices talk about how geo should be integrated into the web in an ethical way?

situx:
A section about ethics has been added to the best practice document.

PR #1386, opened 2022-12-24
remove reference to #1146

PeterParslow:
Because we do now mention ethical use of geo, and #1146 is closed


situx:
Thanks, that seems reasonable

PR #1393, opened 2023-01-20
update coverages section

### LINKED ISSUE #1330: Add illustrations for coverage types
We should have illustrations of different coverage types to let the reader get an impression about the variety of relevant coverage types

lvdbrink:
add examples


situx (commented):
I like the examples, and we could leave it at that, but I wonder if we could also put in an example of a thematic raster image, that is, a raster image that attests meaning to the color ranges of the pixels it represents.
Or is that already covered by one of the examples you included?

Issue #1086, opened 2018-10-24
Spatial Data on the Web - Best Practice 14 - Approach for describing positional accuracy


MichaelGordon:
So far none of the implementation reports ([GNAF](https://github.com/w3c/sdw/blob/gh-pages/bp/BP-implementation-report-00002.md#best-practice-14-describe-the-positional-accuracy-of-spatial-data), [NRW](https://github.com/w3c/sdw/blob/gh-pages/bp/BP-implementation-report-00003.md#best-practice-14-describe-the-positional-accuracy-of-spatial-data), [PDOK](https://data.labs.pdok.nl/publications/SDWBP-Kadaster.html)) implement [BP 14: Describe the positional accuracy of spatial data](https://www.w3.org/TR/sdw-bp/#desc-accuracy). 

Notably it was felt in the GNAF implementation that there was a need for some form of code list to facilitate the discovery of meaning - I believe this view was also shared by @andrea-perego 

Currently there is no detailed discussion of a code list approach in the best practices. This could be an enhancement needed to facilitate implementation.

PeterParslow:
Updated OS Names API implementation report could provide one example of partial conformance to BP 14 (now BP 15), but it doesn't use a code lists approach

PR #1400, opened 2023-03-02
Update OS Names API implementation report

PeterParslow:
OS Names API documentation has moved
Current editors draft introduces a new BP 13, so old BP 13 and on are re-numbered. OS Names API partly conforms to new BP 15 (old BP 14)


PeterParslow:
If this pull request is accepted, updating the existing BP implementation report, then we will have at least on implementation report that states (partial) adherence to old BP 14 / new BP 15 - so we could close #1086

Issue #1286, opened 2021-11-22
[SDW Best Practices Update]: Adding OGC API Features


situx:
Feeding OGC API evolution into the BP. 
An update of the API BP and the "what's wrong with SDI's" section, possibly also other sections.

situx:
We would need some elaboration on the differences between WFS/WCS and OGC API Features somewhere in the Best Practice document.

PR #1405, opened 2023-04-21
remove wfs refs

### LINKED ISSUE #1363: Consider removing references to OGC WFS in the best practice
As discussed in the update meeting on June 30, 2022 the best practice should promote OGC API - Features rather than OGC WFS.  Furthermore, the new generation of OCG APIs are substantially aligned with the best practices and so it may be prudent to mention OGC API - Records, OGC API - EDR, etc. even though many of them are in draft. Perhaps as an aside or prominently in the emerging best practices section.

lvdbrink:
The one large section I'm still removing now is intended. This described how to best expose spatial data if you already have an SDI with WFS services. 

closes #1363

maybe also #1286


situx (approved):

situx:
Seems good to me now. I will merge it in

PR #1414, opened 2023-05-08
Fix member predicate in OWL-Time example

### LINKED ISSUE #1407: Time Ontology - Missing "hasMember" object property definition for modeling proper intervals of hierarchical TRS
Hi,
I'm developing a C# library for giving applications the ability to model ontologies containing temporal entities, so Time Ontology is the conceptual reference I'm following. I noticed, in the geologic time scale example, the usage of a property "**:hasMember**" which is supposed to be an object property from the legacy version of Time ontology (http://www.w3.org/2006/time#) but I cannot find it in the Turtle file.

So my doubt is: how can we specify the proper intervals describing a hierarchical TRS? This property may be missing from the actual draft.

Thanks,
Marco

dr-shorthair:
No `:hasMember` property in OWL-Time.
Use `rdfs:member` instead.

Fixes #1407

### DIFF
--- a/time/rdf/geologicTimeScale.ttl
+++ b/time/rdf/geologicTimeScale.ttl
@@ -193,10 +193,10 @@ geol:Silurian
 geol:TimeScale
   rdf:type :TRS ;
   rdfs:label "Geologic timescale" ;
-  :hasMember geol:Archean ;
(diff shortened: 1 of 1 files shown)


chris-little (approved):
Looks good.

PR #1427, opened 2023-06-21
Introduction section about data ethics

situx:
As recommended by @chris-little here is a small introductory section to spatial data ethics.

Feel free to add and comment.


chris-little (approved):
Good text

situx:
Great! Then I will merge it in!

PR #1438, opened 2024-01-10
[time] Fix russian prefLabel for hour

white-gecko:
The russian prefLabel for hour was `"один час\"@ru"` but should be `"один час"@ru`.

### DIFF
--- a/time/rdf/time.jsonld
+++ b/time/rdf/time.jsonld
@@ -2770,7 +2770,10 @@
       "@language" : "en",
       "@value" : "Hour (unit of temporal duration)"
     },
-    "prefLabel" : [ "один час\"@ru", {
+    "prefLabel" : [ {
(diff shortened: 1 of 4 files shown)


time/rdf/time.jsonld
chris-little, line 1: Seems reasonable, but I am not an expert in the syntaxes and escape characters of ttl, rdf and jsonld as well as which namespaces are supported.
Should the languages tags here for Japanese also be changed `jp` -> `ja`, or is that in the other PR?

white-gecko: Since these are separate concerns, I have separated the changes in two pull requests. I have verified the syntax with rapper/raptor utils resp. the json-ld playground.

dr-shorthair:
Looks like a stray character from some cutting and pasting between apps.

PR #1439, opened 2024-01-10
[time] Fix the language tags of japanese literals to be @ja

white-gecko:


chris-little (commented):
Looks good. but no idea what would break. Probably originally a confusion between ISO 639-1 and ISO3166. And ISO639-2 using `jpn` doesn't help.

white-gecko:
I don't know if something would break, at the moment it is used inconsistently, sometimes ja, sometimes jp.

dr-shorthair:
Shouldn't break anything in term of logic and semantics. These are annotations. Might affect some user-interfaces.

dr-shorthair:
@white-gecko can you resolve these conflicts please?

white-gecko:
I have rebased it.

dr-shorthair (approved):

PR #1451, opened 2025-08-18
Updated WebVMT Editor's Draft

rjksmith:
Added support for negative cue times

PR #1454, opened 2025-08-19
Updated WebVMT Editor's Draft

rjksmith:
Added W3C Unicode styling

### DIFF
--- a/proposals/geotagging/webvmt/index.html
+++ b/proposals/geotagging/webvmt/index.html
@@ -1081,7 +1081,7 @@ mime-type:video/mp4
           Implementations of this specification must not normalize Unicode text during processing.
         </p>
         <aside class='example' title='Unicode Matching'>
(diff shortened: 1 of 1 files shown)


rjksmith:
@TallTed I've added [W3C Unicode code point styling](https://www.w3.org/guide/manual-of-style/#Unicode) as promised in #1451.

Some Unicode characters are not displayed by `&#xxxx;` such as LF and CR, though I believe that this is compliant with the guidance.

Please review and confirm. Many thanks.

TallTed:
I've not reviewed with a fine tooth comb, but the updated source looks OK. I don't have a good way to preview with the PR, but I'll trust you've done as described.

rjksmith:
@TallTed Thanks for your feedback. I'll merge the PR so you can see the result. My Perl regex script dealt with the bulk of the changes and then I tidied up a few anomalies manually.

Please let me know if you have any subsequent concerns. Many thanks for your help.

TallTed:
Unfortunately, there is an update to the template which appears not to have propagated fully yet. The tweak is simply to add a space between `</bdi>` and `<code`, so instead of `</bdi><code`, you will have `</bdi> <code`. I believe this can be done with a global search-and-replace.

